PR #5, opened 2023-02-01
Fix waterswap

chryswoods:
## If this is to fix a bug...

This pull request fixes issue [#426](https://github.com/michellab/Sire/issues/426) in `michellab/sire`.

## Suggested reviewers
@lohedges

This fix is slightly harder to review because there were some autoformatting changes from black, plus I had to clang-format some c++ files that had been missed, plus the "simplified logic" commit was to simplify the way that energy deltas are calculated, as the previous way was broken (this was the "WARNING" that I silenced, and so I fixed thinking this was the original source of the error).

It turned out that the error was very subtle. I've added a unit test so we shouldn't regress. Full details of the fix are in the [commit description](https://github.com/OpenBioSim/sire/commit/696055eb866e7261e54b2a06cc5cf52d6dc87662). The fix is just adding three lines to `wrapper/Mol/__init__.py`

```
@staticmethod
def _molecule_join(views):
    """Join the passed views of a molecule into a single molecule"""
    if len(views) == 0:
        return Molecule()
    # THIS IS THE FIX===
    elif type(views) is not list:
        # this is likely just a molecule itself
        return get_molview(PartialMolecule(views, views.selection()))
    # ==END OF THE FIX
    elif len(views) == 1:
        return views[0]
    else:
        atoms = views[0].selection()
        molnum = views[0].molecule().number()

        for i in range(1, len(views)):
            if views[i].molecule().number() != molnum:
                raise IncompatibleError(
                    "Cannot join different molecules together! %s vs. %s"
                    % (molnum, views[i].number())
                )

            atoms = atoms.unite(views[i].selection())

        return get_molview(PartialMolecule(views[i], atoms))
```

### DIFF
--- a/corelib/src/libs/SireAnalysis/third_party/regress.cpp
+++ b/corelib/src/libs/SireAnalysis/third_party/regress.cpp
@@ -30,174 +30,190 @@
  *
  * first created on December 27, 2007
  * revised on December 30, 2007
-*/
+ */
 
 #include "regress.h"
 
-#include <QObject>
 #include "SireMaths/errors.h"
+#include <QObject>
 
 /*
  * class constructor
-*/
+ */
 Regress::Regress(
-    const list<stREGRESS>& _sample,
-    unsigned int _degree )
+    const list<stREGRESS> &_sample,
+    unsigned int _degree)
 {
-    LoadData( _sample, _degree );
-}   // end of class constructor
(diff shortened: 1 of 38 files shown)


lohedges:
Sorry for not getting to this in time. In any case, I think @nigel-palmer's internal tests are probably a more robust code review :-)

Issue #6, opened 2023-02-06
analyse_freenrg.py is broken


mark-cresset:
The "analyse_freenrg.py" script is currently broken. Around line 520, it tries to get the simulation temperature by parsing the input .dat file. It's assuming that it can split the line on whitespace and get the temperature from field 3 and the unit from field 4. However, the simfile.dat file now reads

#Generating temperature is  298°K

instead of

#Generating temperature is  298 K

so the parsing fails.

In addition, the code later in the script doesn't work as k_boltz now seems to have a bunch of units attached which causes an exception when it is applied to a bunch of matrix slices - I don't understand the units code well enough to work out how to fix this.

**To Reproduce**
Unpack the attached files. Run

sire_python --ignore-ipython analyse_freenrg.py mbar -i simfile_*.dat" -o freenrg-MBAR.dat --temperature 298.0 --percent 95 --overlap

This is with Sire 2023.1.1, Linux

lohedges:
Hi @mark-cresset. Thanks for catching this. As a quick workaround for the `k_boltz` issue you can replace it with `k_bolt.value()`, which will return the value in the given unit.

chryswoods:
Thanks - I'll take a look at the code and will sort out a fix. Thanks for supplying the simfiles. I didn't realise that the mbar part of analyse_freenrg was parsing the raw output files. We'd used s3 files for other tools to make them more robust to formatting changes.

mark-cresset:
Thanks Chris and Lester!

chryswoods:
I've fixed `analyse_freenrg` and confirm that it now works for the simfiles you sent. This is backwards-compatible, so should still be able to parse files produced using older versions of sire. I've submitted the pull request (above) to `devel`. Here's a [link to the changed file](https://github.com/OpenBioSim/sire/blob/fix_analyse_freenrg/wrapper/python/scripts/analyse_freenrg.py).

chryswoods:
There were some residual bugs that needed fixing related to unicode parsing on Windows. I've fixed these, and have added an analyse_freenrg unit test to catch regressions in the future. The fix has been merged into `devel` (so can be used from the latest development binaries on conda-forge) and is now part of `2023.1.2` which is being built now. I think this should be available to download by this afternoon :-)

mark-cresset:
Thanks Chris!

Regards,
Mark

PR #7, opened 2023-02-06
Fixed analyse_freenrg...

chryswoods:
so that it uses `k_boltz.value()` and so that it can cope with `value°unit` temperatures as well as `value unit` temperatures.

The diff is bigger as the script has also been autoformatted using black.

The main fix is changing

```
temp = line.split()[3]
unit = line.split()[4]
```

into

```
temp = line.split()[3]

try:
    unit = line.split()[4]
except Exception:
    # this was a °C or °K (not that °K is correct!)
    temp, unit = temp.split("°")
```

The other change is having kelvin print out as `value K` rather `value°K` - it is not correct to use the degree sign for kelvin (my mistake)

Fix a bug

This pull request fixes issue #6 

## Suggested reviewers:
@lohedges

### DIFF
--- a/corelib/src/libs/SireUnits/convert.cpp
+++ b/corelib/src/libs/SireUnits/convert.cpp
@@ -246,7 +246,7 @@ namespace SireUnits
 
                 strings->insert(DimensionKey(kcal), QPair<double, QString>(kcal, "kcal"));
 
-                strings->insert(DimensionKey(kelvin), QPair<double, QString>(kelvin, "°K"));
+                strings->insert(DimensionKey(kelvin), QPair<double, QString>(kelvin, "K"));
 
                 strings->insert(DimensionKey(degree), QPair<double, QString>(degree, "°"));
 
--- a/wrapper/python/scripts/analyse_freenrg.py
+++ b/wrapper/python/scripts/analyse_freenrg.py
@@ -7,6 +7,7 @@ import warnings
 
 try:
     import sire
+
     sire.use_old_api()
 except ImportError:
     pass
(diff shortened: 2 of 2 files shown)


lohedges (approved):
Looks good to me.

Going forward it would be good to know who is still using `analyse_freenrg` (presumably Cresset). For BioSimSpace, we will be moving over to `alchemlyb` so that we can perform consistent analysis across engines and will deprecate the native analysis tools (`analyse_freenrg` and `gmx bar`) at this point. From then on their will likely be less users of this script, so it would be good to have some tests to ensure that it still works for those who do rely on it.

jmichel80:
I think we will likely still support analyse_freenrg as a native sire app as long as somd is still used. My group members still use analyse_freenrg as the alchemlyb support is not ready yet. In general having a native implementation as a fall back is useful to ensure correctness of the alchemlyb implementation.

lohedges:
Thanks, that's good to know. I do plan on leaving the fallback implementations in BioSimSpace so that it can work if/whenever `alchemlyb` isn't available, or the user doesn't want it in their environment for whatever reason. It would just be good to have a simple test (even just to see if it runs and produces output) in order to catch breakages due to any further Sire API changes going forward. (So we fix it during development, not retroactively.)

chryswoods:
I am planning on creating a regression test in `sire_bigtests` based on the simfiles from Mark. This should give us some level of protection.

analyse_freenrg has two sides to it; there is the side used by Julien for MBAR analysis of somd runs, and there is the side built into sire that is used to analyse free energies from, e.g. waterswap et al. This uses the FEP, TI and BAR objects from SireAnalysis (https://github.com/OpenBioSim/sire/tree/devel/corelib/src/libs/SireAnalysis) to calculate free energies. I think we will always keep both the MBAR and other sides of analyse_freenrg in sire, as (aside from this breakage) they are relatively feature-full, robust and aren't too much to support.

Issue #8, opened 2023-02-14
Unable to read sdf file [BUG]


kexul:
**Describe the bug**

**To Reproduce**
1. Load the files [TAD-03001_minimised.zip](https://github.com/OpenBioSim/sire/files/10730353/TAD-03001_minimised.zip)
2. Run the code 
```python 
import BioSimSpace as BSS
mol = BSS.IO.readMolecules('TAD-03001_minimised.sdf')
```

3. This is the exception that was raised / this is what went wrong.
```
Unable to read the file. Errors are below.



== /data/temp/TAD-03001_minimised.sdf ==

This file could not be parsed by any of the file parsers! It was recognised as a file of type sdf, but all parsers failed to parse this file. The errors from the parsers associated with the suffix sdf are printed below:

*-- Failed to parse '/data/temp/TAD-03001_minimised.sdf' with parser 'SDF'.
The file is not recognised as being of the required format.
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ in <module>                                                                                      │
│                                                                                                  │
│ ❱ 1 mola = BSS.IO.readMolecules('TAD-03001_minimised.sdf')                                       │
│   2                                                                                              │
│                                                                                                  │
│ /data/miniconda3/envs/openbiosim-dev/lib/python3.9/site-packages/BioSimSpace/IO/_io.py:528 in    │
│ readMolecules                                                                                    │
│                                                                                                  │
│    525 │   │   │   │   if _isVerbose():                                                          │
│    526 │   │   │   │   │   raise IOError(msg) from e0                                            │
│    527 │   │   │   │   else:                                                                     │
│ ❱  528 │   │   │   │   │   raise IOError(msg) from None                                          │
│    529 │                                                                                         │
│    530 │   return _System(system)                                                                │
│    531                                                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
OSError: Failed to read molecules from: ['TAD-03001_minimised.sdf']
```
**Expected behavior**
It should not raise an exception. 

 - OS: Linux (centos)
 - Version of Python:  3.9 
 - Version of sire:  latest `dev`

chryswoods:
Thanks for reporting. The issue is that sire's SDF parser is expecting every molecule to end with a separator line (`$$$$`). Because this isn't present, it doesn't think that there are any SDF molecules in the file.

[...]

PR #9, opened 2023-02-14
Feature rdkit

chryswoods:

Pull request to introduce the start of the new sire.convert framework. This has an initial implementation of the sire/rdkit conversion code. This has allowed new functionality to be added, e.g. creation of molecules from smiles strings (`mols = sr.smiles("C1CCCC1")`), conversion back to smiles (`mol.smiles()`), and viewing 2D structures (e.g. `mol.view2d()`).

There is a still a bit to do, but this is worth sharing now so it can be played with by others.

Note that there isn't support yet on Windows because of an issue with missing rdkit headers in their conda package.

I've added `test_rdkit` to test this.

I also added in an `ExcludedPairs` class to handle extra excluded pairs when solving the pmemd bug. This ended up not being used, as there were no missing ExcludedPairs. However, I am including it in case we run into a structure in the future that specifies them.

## If this is to fix a bug...

This also fixes a few bugs. Issues addressed are;

* Issue #8 
* Fixing AmberPrm writing bug when a zero index atom is in position 3 or 4 of a dihedral.
* Fixing agreement of energies between sire and pmemd. Did this by having the `.energy()` function default to the longest cutoff it could.

## Suggested reviewers:
@lohedges

I have more that I want to do for the rdkit code. I will do this in `feature_rdkit2` which will branch from here. I am issuing the pull request now to merge in bug-fixes, plus also to start testing on different platforms. Also, the current code to view 2D structures and generate smiles strings is already very useful.


lohedges:
No worries. I am still plagued by the Windows py39 issue, so know the pain of having everything work locally only to then fail during the CI.

chryswoods:
Issues were (for Linux and Windows) import failures if RDKit was not installed. I've checked and fixed those, plus added in `requirements_test.txt` to install test requirements. For MacOS, the issue was that rdkit requires C++ 2017, and TBB on C++ 2017 wants to use the aligned memory allocators. These aren't supported on MacOS prior to 10.14, which is what is on the runners (or is the target build platform, although I am confused as it works on my Mac with conda-build...). It was a bit of a hack on my part to have most of sire compiled with C++ 2014 and SireRDKit compiled with C++ 2017. So I am now adding a proper C++ 2017 check into the top level cmake file, plus a test for aligned memory allocations. This will then be used to set the flags at the top level, which the rdkit code can then inherit.

chryswoods:
Phew - that all worked :-)

I’ve run sire_bigtests successfully locally, and am happy that I’ve not broken anything.

lohedges:
Great, glad to hear. I'll try to take a look at this later this morning. Can't see there are any issues given that all the tests pass.

lohedges (approved):
Many thanks, this is great. I can confirm that it compiles locally and all BioSimSpace tests pass. At a quick glance through the source code the only thing that I spotted was a couple of duplicate `#include <memory>` directives, e.g. in `majorminorversion.h`. Obviously these don't cause any issues, so feel free to leave them, or fix at a later date.

I'll start a BioSimSpace feature branch to play around with the RDKit conversion behind the scenes, e.g. when setting things up for an MCS calculation, or parameterisation with OpenMM. I imagine that we might get more reliable results now, since we will (hopefully) be able to infer some bond order information that was missing previously when going via an intermediate PDB.

Cheers.

chryswoods:
Thanks - I'll remove them from a future commit. VSCode's global search and replace is great, but doesn't catch things like this ;-)

Another thing it may be good to do is to expose `sire.convert` in `BioSimSpace`, e.g. as something like `BioSimSpace.Convert`? This way we can encourage people to use `BioSimSpace.Convert.to(mol, "sire")` or `BioSimSpace.Convert.to_sire(mol)` rather than use the `mol._sire_wrapper` hidden attribute (and similarly `BioSimSpace.Convert.to(mol)` or `BioSimSpace.Convert.to(mol, "biosimspace")` or `BioSimSpace.Convert.to_biosimspace(mol)` to go back to `BioSimSpace` from `sire`). It may make the code more robust, particularly if we start adding in initialisation or other validation code to the sire wrappers.

chryswoods:
(it would also make the RDKit conversion visible, so Exscientia could load from RDKit directly, and then call `BioSimSpace.Convert.to(rdkit_mol, "biosimspace")` directly)

lohedges:
Yes, that's a good idea. Does the conversion assume a non-legacy Sire System? I guess this is probably what the user will want to play around with externally, but internally I'm still using `sire.legacy.System`. I'll have a play and see.

Agreed, it would be good to directly expose the RDKit conversion too :+1:

chryswoods:
Currently it only works for `MoleculeView`-type (e.g. `Molecule` et al) and `MoleculeView`-container type (e.g. `SelectorMol` et al) objects. The new Sire System is just a thin wrapper around the legacy System, calling the `.molecules()` function to get the `SelectorMol` containing all the molecules. 

I am going to look at converting System objects directly when I write the OpenMM converters. It will be the same public API - just calling a different backend function behind the scenes.

PR #10, opened 2023-02-17
Feature rdkit2

chryswoods:

Adding new functionality

This adds in most of the RDKit code that I wanted to develop. The code from the last push has been refined, functions can now in parallel and more is performed at the C++ level. There is better handling of 2D views. You can also now get smiles and 2D views from molecule containers :-)

I am pushing this now to test the build and run on Linux and Windows. I suspect this will crash at runtime on Windows but need to be sure.

* I confirm that I have added a test for any new functionality in this pull request: [n]

## Suggested reviewers:
@lohedges

I still need to finished off the Stereochemistry class and also add in more documentation and tests. I will do this in `feature_rdkit3`. This merge is mostly to test the link and run, and also to allow the BioSimSpace convert code to be used.


chryswoods:
Thanks goodness this all compiles, links and passes the tests :-)

The root cause of the problem was that the conda build environment is separate from the conda run environment. The build dependencies are installed only in the build environment. RDKit was the first build dependency that I added that was not a compiler or cmake (i.e. a dependency that included a python module and that pulled in other python modules and libraries as dependencies). Installing RDKit into the build environment pulled in lots of other dependencies, which clashed with the ones we needed in the run environment. The clash was because sire's cmake files found the build-environment dependency rather than the run-environment dependency, and so got very confused. Particularly because, once it had found the build environment, it then didn't look in the run environment any more, meaning it couldn't find dependencies that were only in the run environment (like netcdf or TBB). Not RDKit's fault - it was an issue with our cmake files.

Another problem was that the build environment didn't really have python. RDKit was the first python dependency brought into the environment. While Conda-build does pin the python version for the run environment, it doesn't for the build environment (I'd guess because they see no need?). So installing RDKit in the build environment upgraded it to the latest RDKit, in the latest Python (version 3.11). This was incompatible with the libraries in the run environment (python 3.9 or 3.8). This is why I had to add python version pinning to the recipe. I don't think it is needed for the run and host environments, but given we do need it for the build environment, it doesn't hurt to add it to all of them.

And then, on Windows, I ran into another strange issue. I needed to use the `PREFIX` and `BUILD_PREFIX` environment variables to get the absolute file paths to the run environment (`PREFIX`) and the build environment (`BUILD_PREFIX`). These paths use native file separators, e.g. `C:\Users\chris\etc` on windows. This environment variable is read into a CMake variable when building corelib, and is saved in a CMake variables file so that its value can be read by the cmake files that control the wrapper build. However, cmake escapes strings when loading them from its variable file, meaning that the `\U` in `C:\Users\chris\etc` triggers an "invalid escape sequence `\U`" error when the file is read (which was not printed to the GitHub actions log). CMake internally uses Unix separators when saving paths, and is happy with everything being Unix-like, and magically converting to Windows separators when needed (not that this is fool-proof). The solution was thus to read in `PREFIX` and `BUILD_PREFIX` and convert them from Windows to Unix file separators when reading them, so that they would be saved in Unix format.

With this, everything built. I can see from `test_rdkit` that the RDKit C++ library has been loaded correctly on Windows and doesn't crash like it did locally on Windows/ARM. 

It should all now be fine to merge. A plus side is that our build pipeline has been made even more robust!

chryswoods:
I've synched this with `feature_rdkit3` - this completes the Stereochemistry code and also adds in support for Chiralities. I need to add atom hybridisations, and then link across so that this information converts correctly between SDF and RDKit.

lohedges (approved):
Many thanks. I've tested locally and all is fine. The build works and the `BioSimSpace.Convert` tests (which are built on top of this feature) all pass.

Cheers.

PR #13, opened 2023-02-23
Use stack inspection to check if function is called from BioSimSpace sandpit

lohedges:
## This is to fix a bug

Apologies for not spotting this issue. This PR fixes conversion to `BioSimSpace` objects when multiple namespaces are active, as is the case when running `BioSimSpace` tests using `pytest`. In this case both vanilla `BioSimSpace` and `BioSimSpace.Sandpit.Exscientia` are present in `sys.modules`, so calling `sire.convert` functions from regular `BioSimSpace` will return objects of the incorrect type, i.e. those from the Exscientia sandpit. This fixes the issue by using stack inspection to determine the file from which the convert function was called, then working out the sandpit from the file name. (The only thing that I'm not sure of is whether the logic works on Windows, i.e. do the file names use back slashes?) I've also changed things so that the global `_BSS` attribute is updated on each function call. This means things will work if a user changes between sandpits when working interactively, for example.

## Checklist

### DIFF
--- a/src/sire/convert/__init__.py
+++ b/src/sire/convert/__init__.py
@@ -68,8 +68,7 @@ def to(obj, format: str = "sire", map=None):
         return to_openmm(obj, map=map)
     else:
         raise ValueError(
-            f"Cannot convert {obj} as the format '{format}' is "
-            "not recognised."
+            f"Cannot convert {obj} as the format '{format}' is " "not recognised."
         )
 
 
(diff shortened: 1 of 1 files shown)


lohedges:
I think I do need to account for backslash on Windows, so will fix that now.

chryswoods (approved):
This is all good - I will merge this now so that we have the dev packages uploaded for tomorrow morning :-)

PR #14, opened 2023-02-24
Feature rdkit3

chryswoods:

Finishes the RDKit functionality (adding in the Hybridisation object), plus fixes the triclinic_inverse bug (rolls in this fix)

Adding in the Hybridization object so that this info can now be properly preserved and then converted and used by other parts of the code.

## Suggested reviewers:
@lohedges

I've bumped the version to 2023.1.3 based on what we discussed. This functionality is for 2023.2.0, but the version bump should help stop package confusion. No docs for this yet as it falls under the sire.convert code. It will be put into the editing tutorial.


lohedges (approved):
All good. My local build works and all BioSimSpace tests pass.

lohedges:
It looks like no packages were uploaded. Perhaps the storage limit has been exceeded?

chryswoods:
They uploaded, but I hadn't remembered to git tag, so they are still 2023.1.2. I'll git tag and merge the pull request for the extra test I added.

chryswoods:
I've removed the packages that were uploaded, have merged in the extra rdkit tests, have tagged with 2023.1.3, and have pushed this direct to devel to trigger a build.

I think the release process for full releases will be to tag an even-numbered minor number and push the new release, and then, for the first push to devel, to tag an odd-numbered minor number and push the new dev release. This should keep the dev and main version apart (I hope ;-))

lohedges:
I think the existing approach is okay assuming that I don't forget to wait for a new Sire release :-)

Presently the first commit of a new tag (build 0) goes to the `main` _and_ `dev` labels on the channel, then all subsequent builds until the next release go to `dev`. BioSimSpace only has compatibility with the matching Sire release (assuming I update the requirements correctly) so can the `dev` packages can work with Sire builds up until the next. We are constantly testing compatibility, so nothing should break during this time.

One issue that I've noted is that the BioSImSpace `requirements.txt` uses `~=` notation for Sire, which isn't correctly converted into a pinning condition for the conda recipe. This still seems to build okay, but I think it will just pin against the _exact_ version_ (including subsequent builds of this version.) I think this is still fine, and probably what we want anyway, so I might just remove the `~=` bit.

PR #15, opened 2023-03-01
This fixes the ring dihedral issue for the Amber prmtop parser.

chryswoods:
Dihedrals in rings will now have their 1-4 term bit set to ignore, even if they are not listed with the same order in the prmtop file

(e.g. A-*-*-B will be tested against A-*-*-B and B-*-*-A)

## If this is to fix a bug...

This pull request fixes issue reported privately.

## Suggested reviewers:
@lohedges

This doesn't change the energies or the validity (per se) of previous prmtop files produced using sire. The issue is that highly strict prmtop parsers may fail if they strictly require only one 1-4 term per pair of atoms, and they don't check that any subsequent terms are, in fact, equal.

### DIFF
--- a/corelib/src/libs/SireIO/amberprm.cpp
+++ b/corelib/src/libs/SireIO/amberprm.cpp
@@ -2949,7 +2949,14 @@ QStringList toLines(const QVector<AmberParams> &params, const Space &space, int 
             {
                 // NB14 must be computed only the first dihedral found that has this pair,
                 // otherwise we would risk Amber double-counting this parameter
-                const QPair<qint64, qint64> nb14pair(dihs[i], dihs[i + 3]);
+                QPair<qint64, qint64> nb14pair(dihs[i], dihs[i + 3]);
+
+                if (dihs[i] > dihs[i + 3])
+                {
+                    // must ensure a consistent ordering, so that we catch
+                    // A-*-*-B and B-*-*-A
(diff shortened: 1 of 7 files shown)


lohedges:
Ah, very interesting. Do you know if the same issue would occur for our other topology parsers, e.g. `GroTop`? For example, if we converted this file to GROMACS format would OpenMM fail with that too? Perhaps it's just AMBER specific because of the way the 1-4 terms are stored/read.

lohedges (approved):
Looks good. Would it be worth adding a test with the problem topology file to avoid future regression?

chryswoods:
I'll look to see if there would be a similar issue for gromacs. I will also think about a test... (likely have to be grepping for the line in the prmtop file, as sire has corrected the issue on load)

chryswoods:
I've added an extra test that manually checks the prmtop file to ensure that the 1-4 term was written once only. I also looked at the grotop file. This wasn't impacted as it already sorted all of the 1-4 terms (like all other terms) so that they weren't written twice. I added a test to verify this though, including an energy check that it was writing the same system as the prmtop file.

chryswoods:
I've also now added the waterswap fix for Nigel (putting it here to minimise CI/CD). I added a test for this too, so if it passes then all is good to merge.

chryswoods:
I'll merge this in so that the dev packages can be built. We've reviewed the previous parts :-)

PR #16, opened 2023-03-02
Fix preservation of unique properties when updating water topology.

lohedges:
This pull request fixes [this](https://github.com/michellab/BioSimSpace/issues/247) issue. When changing the water topology of a system we need to preserve all molecular properties that are unique to the original molecules, e.g. they might have additional properties to the loaded water template, such as velocities. While there was logic to check for such properties, I was comparing against the _wrong_ molecule when deciding whether to add them.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/corelib/src/libs/SireIO/biosimspace.cpp
+++ b/corelib/src/libs/SireIO/biosimspace.cpp
@@ -210,7 +210,7 @@ namespace SireIO
         // Copy across all properties that are unique to the original molecule.
         for (const auto &prop : molecule.propertyKeys())
         {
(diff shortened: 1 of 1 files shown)


chryswoods (approved):
All good :-)

lohedges:
Thanks. Feel free to consolidate this with the other PR if it will save on CI.

chryswoods:
Thanks - I've pulled this into `fix_openmm_ring_dihedrals` and added some extra tests there as requested in the review. Assuming the tests pass, I will accept the pull request and merge both of these in as a single PR.

PR #19, opened 2023-03-07
Fix BLAS requirement (take 2)

lohedges:
It turns out that we don't need to explicitly link against `libgslcblas` (and in turn `libcblas` and `libopenblas`) in order for Sire to build. Linking against `libgsl` is enough, in which case we implicitly link against `libcblas` and can use any implementation at run time.

The previous issue was that (in a conda environment) `libgslcblas` is a double symlink, so by linking against it you end up depending on whatever blas implementation is present at build time. This means that we have to pin ourselves to an implementation, which restricts the environments into which Sire can be installed.

```
gsl-config --libs
-L/home/lester/.conda/envs/openbiosim/lib -lgsl -lgslcblas -lm
```

`libgslcblas` is a symlink:

```
ls -l ~/.conda/envs/openbiosim/lib/libgslcblas.so
lrwxrwxrwx 1 lester lester 13 Feb  1 14:33 /home/lester/.conda/envs/openbiosim/lib/libgslcblas.so -> libcblas.so.3
```

In addition, this itself is another symlink!

```
ls -l ~/.conda/envs/openbiosim/lib/libcblas.so.3
lrwxrwxrwx 1 lester lester 23 Feb  1 14:33 /home/lester/.conda/envs/openbiosim/lib/libcblas.so.3 -> libopenblasp-r0.3.21.so
```

By using `gsl-config --libs-without-cblas` to specify the linker flags we remove the `libopenblas` requirement from `Sire.Maths`, i.e. we now have:

```
lddtree ~/.conda/envs/openbiosim/lib/libSireMaths.so
...
    libgsl.so.25 => /home/lester/.conda/envs/openbiosim/lib/libgsl.so.25
        libcblas.so.3 => /home/lester/.conda/envs/openbiosim/lib/libcblas.so.3
            libpthread.so.0 => /usr/lib/libpthread.so.0
            libgfortran.so.5 => /home/lester/.conda/envs/openbiosim/lib/libgfortran.so.5
                libquadmath.so.0 => /home/lester/.conda/envs/openbiosim/lib/libquadmath.so.0
            ld-linux-x86-64.so.2 => /usr/lib/ld-linux-x86-64.so.2
...
```
Where previously we had:

```
lddtree ~/.conda/envs/openbiosim/lib/libSireMaths.so                                                   
...
    libgsl.so.25 => /home/lester/.conda/envs/openbiosim/lib/libgsl.so.25
        libcblas.so.3 => /home/lester/.conda/envs/openbiosim/lib/libcblas.so.3
            libpthread.so.0 => /usr/lib/libpthread.so.0
            libgfortran.so.5 => /home/lester/.conda/envs/openbiosim/lib/libgfortran.so.5
                libquadmath.so.0 => /home/lester/.conda/envs/openbiosim/lib/libquadmath.so.0
            ld-linux-x86-64.so.2 => /usr/lib/ld-linux-x86-64.so.2
    libopenblas.so.0 => /home/lester/.conda/envs/openbiosim/lib/libopenblas.so.0
...
```

## Suggested reviewers:
@chryswoods


chryswoods (dismissed):
Looks good - fingers crossed this is the solution :-)

lohedges:
Based on the warning messages from the conda build stage I took the opportunity to rework our requirements. The changes are as follows:

* In addition to the BioSimSpace requirements, we now have `build`, `host`, and `run` requirements files. In particular, the `run` requirements exclude certain packages that aren't needed at runtime, e.g. `tbb-devel` and `libcblas`. (We now have no direct dependency on `blas`.)
* I've switched to using `qt-main` which provides the Qt libraries that we require. (I'm not sure when Qt was split into multiple conda packages.) This will help lower the footprint of our install, since we will no longer pull in `qtconsole` or `qt-webengine`.
* I've removed the manual pinning of the `python` requirement from the `update_recipe.py` script and just added it (unpinned) to the requirements files. With the old approach, conda build was complaining that `python` wasn't present in the run requirements (even though it was), possibly because the pinning rule was malformed. I've checked the logs and a valid python variant was used for all builds on all platforms and the conda build checks now seem happy. (I believe that you needed to add this manual pin since the python version was being changed by `rdkit` during build time on Windows. This doesn't seem to be the case now.)

Now the only warning message is that `rdkit` isn't in the run requirements. However, it is only optionally needed by Sire (which can handle its absence) and pinning is controlled via its presence in the `host` section.

chryswoods (approved):
Thanks Lester - I know that was a lot of fiddly and difficult work!

This feels a lot more logical and robust than what we had before.

lohedges:
I've added a quick fix for detecting water molecules. Somehow I had forgotten to check the dummy atom count in molecules with 4 or 5 atoms, so was wrongly classifying certain molecules as dummies if they had less than 5 atoms and contained two hydrogens and one oxygen. This change is tested through the BioSImSpace test suite, which has been run locally.

chryswoods:
Thanks - feel free to merge once it's built and tests pass.

Issue #21, opened 2023-03-10
Garbage velocities in Gro87 files


chryswoods:
**Describe the bug**
Errors writing Gro87 files when only some molecules have velocities defined.

This is being worked on in the fix_garbage_gro87_velocities

I have fixed the bug. Just need to add a regression test, and will then issue a PR.

I suspect there may be something similar with AmberRst7 that I will double-check too.

PR #20, opened 2023-03-10
Add utility function to remove named property from all molecules.

lohedges:
This PR adds a simple utility function for BioSimSpace that allows the removal of a named property from _all_ molecules in a system. This will make it quick to strip out velocities when combining molecules in any way, e.g. when adding a new molecule to a system.

Since this doesn't touch any of the core Sire functionality I've skipped the build. Feel free to trigger it if you like. Testing and documentation will be done via BioSimSpace. The function call will also be wrapped in a `try/except` block to catch any exceptions in the Python layer.

* I confirm that I have merged the latest version of `devel` into this branch before issuing this pull request (e.g. by running `git pull origin devel`): [y]
* I confirm that I have added a test for any new functionality in this pull request: [y] (Tested via BioSimSpace)
* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: [y] (Documented via BioSImSpace.)
* I confirm that I have permission to release this code under the GPL3 license: [y/n]

## Suggested reviewers:
@chryswoods

PR #22, opened 2023-03-10
Fix garbage gro87 velocities

chryswoods:

Both - this brings in the gro87 fix and feature_removeProperty

## If this is to fix a bug...

This pull request fixes issue #21

This also includes [PR20](https://github.com/OpenBioSim/sire/pull/20) - feature_removeProperty

## Suggested reviewers:
@lohedges

I'll let this run the CI and tests tonight, and will then merge over the weekend if everything is ok (so the packages will then be ready for Monday)


chryswoods:
I checked AmberRst7 and that parser correctly handles the case. But I've just checked AmberRst and that doesn't. I'll add the fix for that too and will re-run the GitHub actions workflow...

chryswoods:
I've added in the AmberRst fix and tests for AmberRst7 and AmberRst. Fortunately there was an internal runtime test that checked for this edge case in AmberRst. The error message is raised wasn't helpful. The fix (which is basically the same as Gro87) means that this runtime test should no longer be tripped.

lohedges (approved):
Looks good to me. Happy for it to be merged when the builds pass.

Cheers.

PR #30, opened 2023-03-24
Remove pymbar from run requirements

chryswoods:
Added ability to add version requirements to `try_import`, so that we can say `pymbar = try_import("pymbar", version="<4")`.

Added that exact code to `analyse_freenrg` so that pymbar is installed at runtime if needed, rather than needing to be installed with sire.

Moved pymbar<4 from a run requirement to a host requirement.

This adds the ability to specify which version of a package should be installed using `sire.utils.try_import` (and `sire.legacy.try_import`). This allows us to install pymbar<4 at runtime, thereby removing the need for it to be in the run requirements. I've moved it to the host requirements.

* I confirm that I have added a test for any new functionality in this pull request: [n - no need as small change]
* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: [n - not a user-facing change]

## Suggested reviewers:
@lohedges

[ci skip]

I'll skip CI as this will be implicitly tested as part of creating the release. I've checked it thoroughly locally and confirm that pymbar<4 is correctly installed. As an aside - pymbar<4 is now available for python3.10 on arm64 (MacOS), so this actually adds functionality for this platform :-)


chryswoods:
The change to `FreeEnergyAnalysis.py` looks bigger because of black autoformatting. The only change is adding these lines to trigger the runtime installation of pymbar

```
    # This should force the installation of pymbar if it isn't
    # installed already
    _pymbar = Sire.try_import("pymbar", version="<4")
```

chryswoods:
Merging across as this should work. I will do a CI/CD on `devel` just to be sure. The sire_bigtest run worked well, so it should be ok. This will let me build and then re-test with sire_bigtest over the weekend, so that I can write the Changelog and do the work to release `2023.2.0` first thing next week

Issue #26, opened 2023-03-16
[BUG] Unable to GROMACS force field parameters from Slipids_2020


ale94mleon:
**Describe the bug**
`bss.IO.readMolecules` is not working as excpected on lipid molecules. 

**To Reproduce**
1. Copy the files in the working directory (they are one POPC lipd and its corresponded GROMACS topology files): 
[files.zip](https://github.com/OpenBioSim/biosimspace/files/10992689/files.zip). I am Using [Slipids2020 force field](https://zenodo.org/record/4194571)
 
3. Run the code

```python
import BioSimSpace as bss
import sire

try:
    membrane = bss.IO.readMolecules(["conf.gro", "topol.top"])
except:
    print(sire.error.get_last_error_details())
```
3. This is the exception that was raised:

> {'where': 'FILE: /usr/share/miniconda3/envs/sire_build/conda-bld/sire_1675934655021/work/wrapper/IO/_IO_load.cpp, LINE: 160, FUNCTION: SireSystem::System load_molecules(const QStringList&, const SireBase::PropertyMap&)', 'why': 'Cannot load the molecules: There is not enough information in this parser (Gro87( title() = Protein in water, nAtoms() = 134, nResidues() = 1, nFrames() = 1, hasCoordinates() = 1, hasVelocities() = 0 )) to start the creation of a new System. You need to use a more detailed input file.', 'pid': '', 'type': 'SireError::io_error', 'backtrace': 'Backtrace disabled. Call sire.error.enable_backtrace_exceptions() to re-enable.', 'from': 'FILE: /usr/share/miniconda3/envs/sire_build/conda-bld/sire_1675934655021/work/wrapper/IO/_IO_load.cpp, LINE: 160, FUNCTION: SireSystem::System load_molecules(const QStringList&, const SireBase::PropertyMap&)'} 

**Expected behavior**
Load the molecule

**(please complete the following information):**
 - OS: Linux (distribution):
 `cat /etc/*-release`:
```bash
LSB_VERSION=1.4
DISTRIB_ID=Arch
DISTRIB_RELEASE=rolling
DISTRIB_DESCRIPTION="Arch Linux"
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://www.archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
[...]

PR #33, opened 2023-03-30
Combined all fixes

chryswoods:
## If this is to fix a bug...

This pull request fixes issue #26, plus the slow AmberPrm issue (fix_amberprm_parallel), plus compile issues on MacOS with older boost (fix_mcs_unary_function) plus the version check on compile (feature_version_check) (which I would call a fix rathe than a feature ;-)).

I have also taken the opportunity to update the GitHub actions workflows and split them into three:

1. pr.yaml - for pull requests only. Not manually runnable. Only builds the pull request for the latest python on Windows, Mac and Linux, and only older Pythons on Linux. This should catch everything except for weird errors on older Pythons on non-Linux systems. It saves 4 out of 9 builds per pull request.

2. devel.yaml - for pushes to devel or manual runs only. Only builds the devel branch for the latest python on Windows, Mac and Linux, and only older Pythons on Linux. This will upload the packages automatically to anaconda cloud. This should catch all but the weird errors, saves 4 out of 9 builds per pull request, and will save space on anaconda cloud as now only the latest Python version will be supported on devel for Mac and Windows.

3. main.yaml - only for manual runs, for which you have to say whether or not you want to upload packages to anaconda cloud (yes / no). This runs everything on the main branch. We can use this to test that fixes / releases do build and pass all tests on all supported platforms and versions, and can choose whether or not to upload the packages based on whether or not we are running a release. Eventually I would like to split this into two, with the run creating packages that are held in GitHub Actions as artefacts, and then the second doing the actual upload if we choose to and the artefacts have been created correctly.

## Suggested reviewers:
@lohedges

I have tested the loading and saving of the SLipids molecules, but can't add any nice tests until I have added in the code to support the forcefield terms in the internal forcefield. I think the code to support reading/writing is a "fix" and can be in 2023.2.1, while the code to add the internal forcefield terms and make sure that these all work correctly (including in openmm) is a "feature" for 2023.3.0. I don't want to block the fix from going in, so it is good to merge. I will do the feature on a proper feature branch.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -25,13 +25,15 @@ jobs:
         exclude:
           # Exclude all but the latest Python from all
           # but Linux
-          - platform.os: macos-latest
+          - platform:
+              { name: "macos", os: "macos-latest", shell: "bash -l {0}" }
             python-version: "3.8"
-          - platform.os: windows-latest
+          - platform: { name: "windows", os: "windows-latest", shell: "pwsh" }
             python-version: "3.8"
-          - platform.os: macos-latest
+          - platform:
+              { name: "macos", os: "macos-latest", shell: "bash -l {0}" }
             python-version: "3.9"
-          - platform.os: windows-latest
+          - platform: { name: "windows", os: "windows-latest", shell: "pwsh" }
             python-version: "3.9"
     environment:
       name: sire-build
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,21 +1,14 @@
-name: Release Main
+name: Pull-Request
 
 # Note that push and pull-request builds are automatically
 # now skipped by GitHub if
 # [skip ci], [ci skip], [no ci], [skip actions], or [actions skip]
 # are in the commit message. We don't need to check for this ourselves.
 
-# Only allow this action to run on a manual run.
-# We should specify when run whether or not we want
-# to upload the packages at the end. We may not want to,
-# as we need to wait for a major or patch release
 on:
-  workflow_dispatch:
-    inputs:
-      upload_packages:
-        description: "Upload packages to anaconda (yes/no)?"
-        required: true
-        default: "no"
+  pull_request:
+    branches: [devel]
+
 jobs:
   build:
     name: build (${{ matrix.python-version }}, ${{ matrix.platform.name }})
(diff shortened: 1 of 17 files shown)


lohedges:
Thanks for this. I agree that my edit is certainly a fix not a feature. Not sure what I was thinking there.

With the changes to the CI I have a question about tags and how we make things persistent. At present we are tagging on `devel` and merging releases across to `main`, so the branches are identical at that point and tags point to the same commit. With the new approach, I'm assuming that fixes get merged into `devel` and `main`, but only `main` gets the patch release tag and the new patch build? If we are going to build patch releases for `main` and `devel` then we would probably need to start tagging `main` and `devel` differently for patch releases, e.g. `2023.2.1` on `main` and `2023.2.1-dev` on `devel`. However, at the point of a full release we will now need to do the work to fix any conflicts when merging `devel` into `main`. We can't simply blat `main` and replace with `devel` since we would lose the commits and tags for the `main` patch releases, which would remove the releases on GitHub. (`conda` packages would be fine, though.)

I'll mirror the same approach for BioSImSpace once we are happy with things.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,76 @@
+name: Pull-Request
+
+# Note that push and pull-request builds are automatically
+# now skipped by GitHub if
+# [skip ci], [ci skip], [no ci], [skip actions], or [actions skip]
+# are in the commit message. We don't need to check for this ourselves.
+
+on:
+  pull_request:
+    branches: [devel]
```
lohedges, line 10: Do we not want to check builds for fix PRs against `main` too? I doubt `main` and `devel` would be so different that the application of the fix would be drastically different between the two branches, but I could imagine cases where something might be subtly different if an new feature has already been applied on `devel` before the fix.

chryswoods: Yes - you are right that I missed this ;-)

chryswoods:
Yes - I haven't quite worked out how tags are going to work. This is the first stage in experimenting to find something thing that is workable. I like the idea of having a 'dev' tag for dev branches, especially as it will then be clearer that dev code is being used. I think main will drift away from devel as we make patch releases, and then we will need to be careful to pull from devel to main and fix any issues / conflicts when we make a main major release.

lohedges:
Great, that makes sense. Was checking that I wasn't misunderstanding anything :-) I don't think conflicts will be too bad. Since we are applying fixes in both places then we should be able to assume the `devel` is the single source of truth at the point of the release, so can simply replace conflicted on `main` files with those `devel` at the point of release. Perhaps missing some edge cases here :thinking: As you say, it will probably take a bit of experimenting to find a process that works.

PR #34, opened 2023-04-01
fix_combine_all into main

chryswoods:
This is a PR that brings all the latest `fix_` branches that have been accepted into `devel` and ports them into `main`. In this case it is a simple PR as the only change from `devel` since the last sync point was this PR :-). Future PRs will be more complex as they may require some back-porting.

I've turned on the PR GitHub Action for PRs into `main`, so this should be automatically built and tested. @lohedges - if the test passes could you double check that this is doing the right thing? There's no rush as this code won't be built into a package until we do the `2023.2.1` patch release (end of April?).

## Suggested reviewers:
@lohedges

### DIFF
--- a/.github/workflows/devel.yaml
+++ b/.github/workflows/devel.yaml
@@ -1,21 +1,15 @@
-name: Release Main
+name: Release Devel
 
 # Note that push and pull-request builds are automatically
 # now skipped by GitHub if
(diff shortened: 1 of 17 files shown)


chryswoods:
Merging so that I can test the pinning with the newest version (in case I make a 2023.2.1 release)

Issue #42, opened 2023-04-11
[BUG] AmberPrm parser segfaults if RESIDUE_LABEL flag is empty


xiki-tempula:
**Describe the bug**
Sorry for pushing this bug to you. It is a interchange 0.3 issue. Do you mind chat with the interchange people to fix this issue? Thanks.

**To Reproduce**
Create a new environment `mamba create -n test -c conda-forge -c OpenBioSim/label/dev biosimspace=2023.2 AmberTools`
```
import BioSimSpace as BSS
BSS.Parameters.parameterise("c1ccccc1C(=O)[O-]", "openff_unconstrained-2.0.0").getMolecule()
```
Gives
Segmentation fault (core dumped)

**Expected behavior**
Runs

**(please complete the following information):**
 - OS: Linux (distribution)
 - Version of Python: 3.8
 - Version of BioSimSpace: 2023.2
 - I confirm that I have checked this bug still exists in the latest released version of BioSimSpace: [yes/no]

**Additional context**

lohedges:
For reference, it looks like they have recently made some changes to the logic to generated residue labels [here](https://github.com/openforcefield/openff-interchange/commit/6025b546c9fa96ae0d1dd763fcf31be56f559b46). I assume that this isn't working correctly. I'll create a minimal example and raise an issue there.

lohedges:
I've now reported the issue at the link above.

chryswoods:
It was an easy fix - test added, pull request submitted and we'll wait to see if it all works before accepting into `devel` / `main` :-)

lohedges:
This has now been fixed in OpenFF and will appear in version `0.3.1` within the next few weeks. This should mean that older builds of Sire should work again.

xiki-tempula:
Great. Thanks. So Sire 2023.2.2 would work with openff-interchange 0.3?

xiki-tempula:
Great thanks

PR #43, opened 2023-04-12
fix_42

chryswoods:
Fixed the segfault when there is no residue label (and also the segfa…ult that would have occured if there was no atom label). Added a unit test to check.

## If this is to fix a bug...

This pull request fixes issue #42

## Suggested reviewers:
@lohedges

### DIFF
--- a/corelib/src/libs/SireIO/amberprm.cpp
+++ b/corelib/src/libs/SireIO/amberprm.cpp
@@ -3730,12 +3730,32 @@ MolStructureEditor AmberPrm::getMolStructure(int molidx, const PropertyName &cut
     // the layout of cutgroups, residues and atoms
     MolStructureEditor mol;
 
-    const auto atom_names = this->stringData("ATOM_NAME");
+    auto atom_names = this->stringData("ATOM_NAME");
+
+    while (atom_names.count() < natoms)
+    {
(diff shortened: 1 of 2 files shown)


lohedges:
Thanks for this. I guess there is still a potential issue if there are missing atom/residue labels for a multi atom/residue system. Since we only append `UNK` at the end to match the length of the corresponding pointers array `UNK` would correspond to the wrong atom/residue if we were missing a label from the middle. I imagine this situation is so unlikely that it can just be ignored, and I'm not sure what problem it would cause anyway. (Other than waters, which need precise names.)

chryswoods:
I agree - my rationale was that that situation would only arise for a malformed / broken input file. I could raise an exception rather than "fixing" it, which to be honest I was very tempted to do. But I think this is such an edge case that (hopefully) a lot of "UNK" atom/residue names would flag something as strange to the user, or they would know what they are doing, and I didn't want to force them to fix their file to get it to load.

It was a decision guided by the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) - I'll liberally accept empty names, but will conservatively output meaningful names (`UNK`).

chryswoods:
This is included in `fix_44` and will be merged in with that.

PR #45, opened 2023-04-13
fix_issue_44 (AmberPrm parallel segfault)

chryswoods:
Here is the updated AmberPrm that fixes the segfault. The issue was that when I originally wrote it, I was not sufficiently careful to access Qt container data using raw pointers (especially const raw pointers where this was needed). This meant that the Qt container infrastructure was getting confused with the reference counting, plus I was occasionally creating modified copies of data that were double-freed.

This fix updates the code to use `.data()` and `.constData()` to get the raw pointers to the underlying containers, plus adds const references where needed. I have tested this using the script in the issue, and have run up to 200,000 loads and saves without crashing, in full parallel.

## If this is to fix a bug...

This pull request fixes issue #?

* I confirm that I have merged the latest version of `devel` into this branch before issuing this pull request (e.g. by running `git pull origin devel`): [y/n]
* I confirm that I have permission to release this code under the GPL3 license: [y/n]

* I confirm that I have merged the latest version of `devel` into this branch before issuing this pull request (e.g. by running `git pull origin devel`): [y/n]
* I confirm that I have added a test for any new functionality in this pull request: [y/n]
* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: [y/n]
* I confirm that I have permission to release this code under the GPL3 license: [y/n]

## Suggested reviewers:
@chryswoods, @lohedges

### DIFF
--- a/corelib/src/libs/SireIO/amberprm.cpp
+++ b/corelib/src/libs/SireIO/amberprm.cpp
@@ -161,27 +161,35 @@ void AmberPrm::rebuildExcludedAtoms()
                                   CODELOC);
     }
 
+    const auto nnb_per_atom_data = nnb_per_atom.constData();
+    const auto inb_data = inb.constData();
+
     excl_atoms = QVector<QVector<QVector<int>>>(nmols);
+    auto excl_atoms_data = excl_atoms.data();
     int nnbidx = 0;
 
     // we need to loop through the excluded atoms in atomnum order, not molidx order.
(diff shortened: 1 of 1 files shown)


lohedges (approved):
Many thanks for fixing this so quickly. Glad it wasn't too painful, i.e. anything to do with the actual logic of reading/writing.

PR #55, opened 2023-05-19
Feature s3 header

chryswoods:
## This introduces new functionality...

This pull request adds the ability to set set arbitrary metadata that will be added to all s3 files written by sire. I have also exposed the sire.legacy.Stream functions that are useful in sire.stream, fixing some rough edges.

The arbitrary metadata is stored internally in the `FileHeader` object, as a `QHash<QString,QString>`. Any picklable type is pickled and then hex-encoded and passed in as a `QString`. The reverse process occurs when reading a property. The properties are set globally, so that they are saved to any subsequently written s3 file.

The `test_stream_header.py` file shows how to use the API. I have not added anything more than API docs as I don't think this should be widely advertised (it is more to better-support BioSimSpace's stream functionality).

You set a global header property using `sr.stream.set_header_property(key, value)`, and can clear it by setting it to None, e.g. `sr.stream.set_header_property(key, None)`.

As this can be any picklable type, you could create a python metadata object that contains all the extra data you want, and then set a single global header property, e.g. `sr.stream.save_header_property("my_metadata", my_metadata_object)`.

You can then get this data from the loaded `FileHeader`, e.g. via

```
>>> h = sr.stream.get_data_header(s3file)
>>> my_metadata_object = h.property("my_metadata")
```

This will return `None` if the metadata hasn't been set. You can also use `h.has_property("my_metadata")` to query for it.

Note that this increases the global s3 version number from 2 to 3. This means that s3 files written by this version of sire CANNOT be read by earlier versions. But, files written by earlier versions can still be read (it is backwards compatible). s3files from earlier versions of sire don't have any set properties.

* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: [n - as explained above]

## Suggested reviewers:
@lohedges

### DIFF
--- a/corelib/src/libs/SireStream/streamdata.cpp
+++ b/corelib/src/libs/SireStream/streamdata.cpp
@@ -392,7 +392,11 @@ namespace SireStream
     SIRESTREAM_EXPORT void setHeaderProperty(const QString &key, const QString &value)
     {
         QMutexLocker lkr(detail::dataMutex());
-        header_properties[key] = value;
+
+        if (value.isEmpty() or value.isNull())
+            header_properties.remove(key);
+        else
+            header_properties[key] = value;
     }
 
     SIRESTREAM_EXPORT QString getHeaderProperty(const QString &key)
--- a/corelib/src/libs/SireStream/streamdata.cpp
+++ b/corelib/src/libs/SireStream/streamdata.cpp
@@ -62,8 +62,8 @@
 
 using namespace SireStream;
 
-using std::shared_ptr;
 using boost::tuple;
+using std::shared_ptr;
 
 namespace SireStream
 {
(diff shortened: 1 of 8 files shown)


lohedges (approved):
I've tested locally and everything is working fine. On the BioSimSpace side I had to make a few small adjustments to handle streaming via `sr.stream` rather than `sire.legacy.Stream`, i.e. using `sr.system.System` rather than `sire.legacy.System`, then converting back on load. When I get time, I'll update the BioSimSpace object constructors so that it can support the new `System` object too. (This just involves updates in more places.)

chryswoods:
I half hate myself that I made this new System class ;-). It has become a bit of a special case... I made it back at the beginning of the modernisation work before I'd seen how powerful monkey patching and the new SelectorMol et al classes could be. I'll still keep it, but can see it being something that will be revised in the hypothetical future when I have free time ;-)

Issue #58, opened 2023-05-24
[BUG] SMILES conversion can return an empty molecule


lohedges:
I assume this is because the SMILES conversion is failing, but it's possible to get back an empty molecule, e.g:

```python
In [1]: import sire as sr

In [2]: sr.smiles("c3cc[c+]2cc(C1CCCC1)[nH]c2c3")

[13:03:23] Can't kekulize mol.  Unkekulized atoms: 0 1 2 4 5 12 13
Out[2]: Molecule( :0      num_atoms=0 num_residues=0 )
```

If an exception was raised by `rdkit` then it would be good if this was caught. From working with BioSimSpace, I know that `rdkit` often doesn't raise an exception and just returns nothing, so it might be necessary to check for this too.

If I go via `rdkit` directly, then I see the following:

```python
In [1]: from rdkit import Chem

In [2]: rdmol = Chem.MolFromSmiles("c3cc[c+]2cc(C1CCCC1)[nH]c2c3")
[13:05:40] Can't kekulize mol.  Unkekulized atoms: 0 1 2 4 5 12 13

In [3]: type(rdmol)
Out[3]: NoneType
```

chryswoods:
I'll take a look at this in the `fix_58` branch.

chryswoods:
I've raised a pull request with the fix. This adds a `must_sanitize` flag (default True) to `sr.smiles()`. We now get an exception if the molecule doesn't sanitize. It is currently a `ValueError`, but I am open to changing this (I struggle choosing Python exceptions as I don't want to create new ones that are specific to sire, but many of the default exceptions also feel wrong).

If `must_sanitize` is False, then we now run sanitization manually, skipping steps that individually fail. This results in a molecule, but the coordinates and 2D structure are pretty poor...

PR #62, opened 2023-05-30
Fix RDKit null molecule issue (58)

chryswoods:
Added a `must_sanitize` option that forces full sanitization of molecules generated

from smiles strings, and raises an exception if they fail to generate. This is set as default True.

If this is False, then sanitization is run manually, skipping any stages that fail.

Added a unit test to check this, plus added an addition to the documentation explaining the new mode.

## If this is to fix a bug...

This pull request fixes issue #58

## Suggested reviewers:
@lohedges

We are only a month away from the next major release. Do you think we are close enough that we can hold this for `2023.3.0` and not back-port to `2023.2.X`?

### DIFF
--- a/doc/source/tutorial/part05/03_smiles.rst
+++ b/doc/source/tutorial/part05/03_smiles.rst
@@ -72,6 +72,15 @@ create an `rdkit Molecule <https://www.rdkit.org/docs/source/rdkit.Chem.rdchem.h
 which is converted to a :mod:`sire` :class:`~sire.mol.Molecule` using
 the functions in the :mod:`~sire.convert` module.
 
+.. note::
+
+   Note that the default is that all sanitization steps provided by rdkit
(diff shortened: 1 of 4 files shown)


lohedges:
Yes, I'm happy to delay if you think a full fix is fiddly. In the first instance the main thing is just you catch the error so we don't end up with an empty molecule in BioSimSpace during parameterisation. If it's easier, for now I could just catch this at my end.

chryswoods:
This is a full fix - it was more me being lazy and trying to skip the back-port ;-)

I can back-port it if you think we will make another patch release before 2023.3.0.

chryswoods:
I've just synchronised this with `devel`, and cancelled the CI as this PR is included in `feature_trajectory_save`.

chryswoods:
Closing as this is automatically included in `feature_shared_properties`

PR #66, opened 2023-06-02
Feature trajectory save

chryswoods:
This PR (finally) includes all of the work to properly support trajectories in sire. This moves all the trajectory parsers onto a common framework and base, makes them all use the Trajectory and Frame classes properly, and adds proper trajectory save support to all trajectory parsers.

Major internal changes are that all trajectory parsers now stream frames from their files. This means that files cannot be changed while they are being read, as otherwise undefined behaviour will result.

As part of this, I have expanded the number of file formats we support. We fully support Gromacs TRR and XTC formats, and now have a complete DCD parser.

Another feature included with this is a fully working ProgressBar class. This is a C++ class that interfaces back to Python so that it can print to the screen, to a ipython session, or to a Jupyter notebook in a thread-safe and GIL-compliant manner. There is a new `SireBase::release_gil()` function that lets sire communicate to Python when it can release the GIL. And there are hooks to let sire write to Jupyter and to `sys.stdout` which properly re-acquire the GIL. This is all massively useful, as now I can create progress bars in C++ and not have to do jumping back and forth within Python (e.g. chunking up C++ function calls). You can see the new progress bar in the saving functions in the parsers. I have also updated the dynamics, minimisation, and trajectory analysis (e.g. energy calculation) code to use this new progress bar class and to not have to chunk via Python.

A further benefit is that CTRL+C or Jupyter interrupt can now stop long-running and multi-threaded C++ functions. This is because there is a now a new SIGINT signal handler in ProgressBar that checks for SIGINT and that will trigger an interrupt exception when the C++ function next updates the progress bar.

Finally, I added in support for reading and writing single-frame file formats as trajectories, by reading and writing to directories that contain "frame" files. The frame number and system time are encoded into the filename.

## Suggested reviewers:
@lohedges

This PR includes [Fix RDKit null molecule](https://github.com/OpenBioSim/sire/pull/62), just to save on CI/CD.

The trajectory code is documented in a detailed guide (cheatsheet/trajectory.rst) and a tutorial page (part01/06_trajectories.rst). The tests are in test/io/test_trajectories.py, which saves and loads a trajectory in each of the supported formats.

I have also tested this manually by loading trajectories from the MDAnalysis test set. I am thinking about how we can add these to, e.g. sire_bigtests for some longer regression tests before each release.


chryswoods:
I added some fixes for Windows and also pulled in the change I made to devel to disable to boresch tests if scipy isn't installed.

Issue #59, opened 2023-05-25
feature_somd_freenrg_positional_restraints


jmichel80:
This feature branch contains code that generalised the positional restraints and constraints functionality available in somd-freenrg. 

**Is your feature request related to a problem? Please describe.**

The current code offers limited options to control which atoms in a system are subject to positional restraints. Currently it is possible to appy positional restraints to every heavy atoms present a system, excluding atoms contained in a list of excluded residue names, as described [here](https://github.com/OpenBioSim/sire/blob/devel/wrapper/Tools/OpenMMMD.py#L1020-L1037). 

It is only possible to freeze atoms that are contained in a list of allowed residue names, as described [here](https://github.com/OpenBioSim/sire/blob/devel/wrapper/Tools/OpenMMMD.py#L1194-L1215)

**Describe the solution you'd like**

It should be possible to restrain or freeze some or all atoms present in a list of residues. Such list of atoms would be contained in a csv file autogenerated by a preprocessing script run on the prm7/rst7 files used as input to somd-freenrg. The format would look like

```
#ResidueNumber, #AtomNumber
1,all
2,45-54
8,all
10,185-199
```

The residue specific positional restraints would be activated by specifying a config file keyword option 

`positional restraints = path_to_restraint_file `

lohedges:
Thanks, @jmichel80. Just a few thoughts / comments...

> The format would look like
>#ResidueNumber, #AtomNumber
>1,all
>2,45-54
>8,all
>10,185-199

Perhaps this is convenient for the intended users interfacing with SOMD from the command-line, but I was wondering if there is a reason why the restraints couldn't be provided as a list of atom indices? I only ask because this is how the current position restraint system in BioSimSpace works, and is how we currently interface with every other engine. It is obviously possible for us to convert atom indices to the format that you propose, but it is a bit more work. Using atom indices would also allow the restraints to be easily entered in the configuration file directly, e.g. as a comma-separated list of indices. That way the configuration option could be an explicit list of restraints, or a path to a restraint file. When just using atom indices you also don't need to worry about parsing both strings and integers, and index regions in different formats, e.g. `45-54`, `46,47,49-52,56`, etc. (Although we could still support compact range formatting.) I guess we could support both approaches (atom indices and residue number plus atom number), or provide a utility script/function within BioSimSpace to convert formats. (There is already code in BioSimSpace that would easily convert your suggested format to absolute atom indices, but not the reverse.)

[...]

Issue #49, opened 2023-04-26
[BUG] Fixed-precision causes different triclinic box angles during triclinic lattice reduction


xiki-tempula:
**Describe the bug**
Solvate a system with rhombic dodecahedron square box would give a non-integer degree

**To Reproduce**
```
import BioSimSpace.Sandpit.Exscientia as BSS
mol = BSS.Parameters.openff_unconstrained_2_0_0("C").getMolecule()
solvated = BSS.Solvent.tip3p(mol, box=[4*BSS.Units.Length.nanometer]*3, angles=[120*BSS.Units.Angle.degree, 120*BSS.Units.Angle.degree, 90*BSS.Units.Angle.degree])
solvated.getBox()[1][0].value()
```
gives 119.99998318630921

**Expected behavior**
gives 120 or at least to the precision of double

msuruzhon:
Thanks @lohedges and @chryswoods, it would be great if we could also have some temporary quick workaround in the current version of Sire or BSS as well, as this bug means we can't realiably run dodecahedral REMD in our current release and somehow that didn't use to be an issue before (not sure why).

lohedges:
Yes, it's strange that the issue hadn't occurred before. I'm not sure how you are setting things up, so perhaps this isn't possible if the replicas are initialised from different Python processes but you could do something like the following to set the box angles to the same value for all replicas:

```python
# Here replicas is a list of the systems, one for each replica.

# Get the box dimensions and angles for the first replica.
box0, angles0 = replicas[0].getBox()

# Apply the same angles to each other replica, keeping the box dimensions.
for replica in replicas[1:]
    replica.setBox(replica.getBox()[0], angles0)
```

For example, using the input that @xiki-tempula posted in [this](https://github.com/OpenBioSim/biosimspace/issues/55) thread:

```python
import BioSimSpace as BSS

mol_0 = BSS.IO.readMolecules(['lambda_0.0000/amber.prm7', 'lambda_0.0000/amber.crd'])
mol_1 = BSS.IO.readMolecules(['lambda_1.0000/amber.prm7', 'lambda_1.0000/amber.crd'])

# Initial boxes.
mol_0.getBox()
([81.5793 A, 81.5793 A, 81.5793 A],
 [60.0000 degrees, 60.0000 degrees, 90.0000 degrees])
mol_1.getBox()
([81.4219 A, 81.4219 A, 81.4218 A],
[...]

PR #68, opened 2023-06-07
Fix re-occurence of issue #49

lohedges:
This PR fixes the re-occurence of issue #49 reported in #59. By changing the direction of the rounding bias we achieve stable angles _and_ a reduction that OpenMM is happy with. It is also possible to use a _smaller_ bias (smaller than the accuracy of the coordinates, angles, and box dimensions in the input files) which makes the reduction more precise. Another consequence is that our reduction now matches the default box angles suggested by GROMACS [here](https://manual.gromacs.org/current/reference-manual/algorithms/periodic-boundary-conditions.html). (Although the previous ones were a completely equivalent space.)

I've tested NPT simulations with each engine that BIoSimSpace supports and the box angles are stable when getting back a system at all points along the trajectory. I've also added a test to make sure that OpenMM remains happy with the reduction we apply to the standard triclinic space used by Cresset.

Fingers crossed this does the job :crossed_fingers: 

## Suggested reviewers:
@chryswoods


chryswoods (approved):
All looks good. Hopefully this will work.

I've updated it to the latest changes in devel. Feel free to merge when you are happy :-)

chryswoods:
I'll merge this - I am about to submit the units PR which will roll up everything else. Merging your PR will make it easier to submit mine :-)

lohedges:
Great. We'll just have to make sure that `devel` is merged into @mb2055`s branch before merging so that my fix overwrites the temporary workaround.

chryswoods:
I'll merge in @mb2055's branch after merging in the (monster) PR I am about to submit ;-). This will merge everything together.

Issue #63, opened 2023-05-31
[BUG] How to correctly calculate the center of mass for wrapped molecules


lohedges:
In trying to fix [this](https://github.com/OpenBioSim/biosimspace/issues/86) issue I'd like to translate a molecule (or molecules) back to the box center based on their current center-of-mass. Doing this via the Sire object evaluator doesn't work if the molecule is wrapped across the boundary of any periodic box since the calculation doesn't take account of the space.

It is possible to fix this in BioSimSpace by manually computing the center of mass as shown below, but this is slow for large systems. Is it possible to get the evaluator to respect the space, or should I just add a C++ utility function to do this for me, e.g. like the other BIoSimSpace ones.

Here is a unit test that reproduces the issue:

```python
from math import isclose as _isclose
from sire.legacy.Maths import Vector as _SireVector

import BioSimSpace as BSS


def test_com_wrapping():
    # Load the single-molecule system.
    system = BSS.IO.readMolecules("wrapped.*7")

    # Compute the centre-of-mass using Sire's evaluator.
    com_sire = system[0]._sire_object.evaluate().centerOfMass()

    # Now compute it manually, taking into account periodic boundaries.
    # We want the centre-of-mass of the molecule as a connected object.

    # Store the space.
    space = system._sire_object.property("space")

    # Get the first atom in the molecule..
    atom = system[0].getAtoms()[0]

    total_mass = atom._sire_object.property("mass").value()

    # Reference coordinate.
    ref_coord = atom._sire_object.property("coordinates")

    # Initialise the centre-of-mass.
    com_bss = total_mass * atom._sire_object.property("coordinates")

    # Sum over the rest of the atoms.
    for atom in system[0].getAtoms()[1:]:
        # Get the mass and update the total.
        mass = atom._sire_object.property("mass").value()
        total_mass += mass

        # Get the coordinate and add to the reference coord using the
        # its distance from the reference in the minimum image convention.
        coord = atom._sire_object.property("coordinates")
        coord = ref_coord + _SireVector(space.calcDistVector(ref_coord, coord))

        # Update the centre of mass.
        com_bss += mass * coord

    # Normalise.
    com_bss /= total_mass

    # Make sure the centre-of-masses are approximately the same.
    for x, y in zip(com_sire, com_bss):
        assert _isclose(x.value(), y.value())
```

The input files are  [here](https://github.com/OpenBioSim/sire/files/11615676/wrapped.tar.gz). 

[...]

PR #71, opened 2023-06-08
Feature shared properties

chryswoods:
This pull request fixes issue #63 

This also introduces new functionality.

I've added two useful things:
    
1. Shared properties on System. These are shared properties which will be
   copied (and kept up to date) to all contained molecules.
   Use `system.add_shared_property(name, value)` to add a shared
   property. By default, space and time are shared.
    
2. I've add a `make_whole` function to `mol.move()`. This wraps
   a molecule into a space, making sure that it is not broken
   across periodic box boundaries. This is exposed both to
   Cursor, and also to System, so that you can now write
   `system.make_whole()` to make all contained molecules
   whole.

The shared property code makes it easy to call `make_whole` 
wherever a molecule is, as we can nearly always be sure
that it will have a valid space.

The `make_whole` function is not called automatically on load,
but is easy enough to call after you have loaded a system.
Just use `mols.make_whole()` on a system, or use
`mols = mols.cursor().make_whole().commit()` on any other 
molecule view / molecule container.

This function will wrap into the box based on the current center
of coordinates (either the centre of the system, or the 
centre of molecule, depending on what is made whole).

You can use an optional `map` parameter to specify which properties to use for the coordinates and space (including passing in a space yourself).

## Suggested reviewers:
@lohedges

This builds on top of the monster PR ;-). Accepting this will accept the rest.

I've not added any documentation (beyond API) as I am not sure how visible I want shared properties to be yet. I am also not sure where to put `make_whole` in the tutorial. I suspect I will eventually have a detailed movement page (like the search page, or units page).


chryswoods:
I have merged in @mb2055's pull request to `devel`, and have merged that into this PR. I have fixed the test failure, which was caused by other tests changing the `ala_traj` fixture. The failing test had assumed (as it should) that it was working with an unaltered `ala_traj`. It was all my fault ;-)

I am thinking of adding another checkbox to the PRs saying something like "I confirm I have re-run all of the unit tests in "veryslow" mode and these pass [y/n]". This would have saved the CI. I had mistakenly only run the new tests individually...

chryswoods:
I've optimised the `make_whole` code and now added a map option (`map={"make_whole": True}`) that can automatically make molecules whole on load (`sr.load`) and when loading a trajectory frame (`mol.load_frame`). This is default `False` if not set, so as not to change existing behaviour.

The optimised code takes microseconds to small milliseconds to make a system of molecules whole.

PR #72, opened 2023-06-14
Feature polish

chryswoods:
This pull request contains code that cleans up sire before the 2023.3.0 release.

* We have fixed compile issues related to GCC 13.
* Optimised the load and save routines, so they are much faster
* Optimised other parts of the code, e.g. converting to openmm
* Made some functions easier to call by automatically converting strings to units, matching file formats ignoring case, matching openmm platforms ignoring case, and using OpenMM-metal if that is installed on MacOS.

## Suggested reviewers:
@lohedges

I am submitting this to test that things are now fixed on GH Actions. I've tested locally on Mac and Linux, including running big_tests. Please check with BioSimSpace (I will check too). You should notice the speedup for reading AmberPrm files ;-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,5 @@ tests/cache/*
 #VS Code
 .vscode
 
+.DS_Store
+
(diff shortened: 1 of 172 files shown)


chryswoods:
I found the cause of the BioSimSpace failures - I had broken the writing of excluded atoms in the optimisation to just write them for waters (rather than calculate them). I had forgotten to add `start_idx` to the indexes, and had also forgotten to only write out pairs for `i < j`. This commit fixes this. I will submit the BioSimSpace PR too. Between the two of these, I am hoping that everything should be working ;-)

lohedges (approved):
All good. Everything passes locally.

PR #74, opened 2023-06-21
Feature smarts

chryswoods:
This PR adds smiles and smarts search support. You can now search container using smiles or smarts strings. These use the C++ `RDKit::SubstructMatch` under the hood, returning *all* matches via new `AtomMatch` and `AtomMatchM` classes.

I have added tests, plus added documentation via additions to the "search" detailed guide, plus a new tutorial page (part05/04_smarts).

## Suggested reviewers:
@lohedges 

This is it now for `2023.3.0` (really!). I have filled in the changelog ready for the release. Please do add anything that I have missed via a [ci skip] commit. @mb2055 - please could you add a changelog entry for your recently accepted PR?

(I am thinking we should probably update the changelog at PR rather than at release... ;-))


chryswoods:
I suspect this will fail with a link error on Linux and possibly a compile error for the wrappers on Windows. I will fix these tomorrow ;-)

(the Mac is far more permissive when compiling...)

lohedges:
Yep, I get the same issues when compiling locally. Let me know when you've fixed and I'll test at my end.

lohedges:
Great, that works :+1:

chryswoods:
Windows is such a pain...

It has failed to recognise that templates instantiated in libSireMol are the same as those in _Mol, and then de-duplicate them (as should normally be the case). All the help I've found so far starts with "Windows is a pain" and the threads are random rants about the weird ways visual studio c++ handles templates...

I'll fix it, but it is a try - long compile - test kind of problem...

chryswoods:
Ok - I refactored `AtomMatch` and `AtomMatchM` so that they don't inherit from a template class (`Selector<Atom>` and `SelectorM<Atom>`). They still have all the functionality because I either chose a non-template class (`PartialMolecule`) or mirrored the functions so it can duck-type.

This now compiles, links, runs and passes all tests locally on MacOS, Linux and Windows. Let's see what it does on GH Actions...

(this also now includes the minor fixes I mentioned in #75)

PR #78, opened 2023-07-13
Fix AmberRst bug (wrong velocities)

chryswoods:
Fixed the bug in AmberRst, where velocities were written out in units of A ps-1, but needed to be written as AKMA time velocities.

Also added in the labels as missing these caused issues for cpptraj

This is the backport into main.

* I confirm that I have merged the latest version of `main` into this branch before issuing this pull request (e.g. by running `git pull origin main`): [y]

This is a quick PR to fix Exscientia's deployment.

### DIFF
--- a/actions/update_recipe.py
+++ b/actions/update_recipe.py
@@ -201,6 +201,7 @@ def check_reqs(reqs0, reqs1):
 # check_environment_reqs(env_reqs)
 
 build_reqs = dep_lines(check_reqs(build_reqs, env_reqs))
+host_reqs = combine(host_reqs, bss_reqs)
 host_reqs = dep_lines(combine(host_reqs, env_reqs))
 run_reqs = dep_lines(check_reqs(run_reqs, env_reqs))
 test_reqs = dep_lines(check_reqs(test_reqs, env_reqs))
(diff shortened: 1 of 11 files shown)


xiki-tempula:
Thank you very much for the help. When you are testing the fix, do you mind ensure that if you load in `amber.crd` and `amber.prm7` and then output to RST7 format, you would get something that is similar to `amber.rst`, please? I generate the `amber.rst` from `amber.crd` with cpptraj. Thank yo
[Archive.zip](https://github.com/OpenBioSim/sire/files/12049447/Archive.zip)
u.

chryswoods:
I've added that as a unit test. The velocities are now all good. The triclinic angles are flipped though... Here's the diff

```
1c1
< Cpptraj Generated Restart                                                       
---
> triclinic_protein
38503c38503
<   81.4008827  81.4008827  81.4008989 119.9999934 119.9999934  90.0000000
---
>   81.4008827  81.4008827  81.4008989  60.0000066  60.0000066  90.0000000
```

(I renamed the file to 'triclinic_protein' for the test)

chryswoods:
This contains a number of fixes that are going into main

* AmberRst velocities are written with the wrong scale factor
* DCD trajectories are written with the wrong header
* Legacy sire scripts print outputs with only base units
* Conda recipe had dropped the BSS deps so wouldn’t build without an environment file.

These changes will go into devel with another PR. They’ve gone into main first as they are needed for 2023.3.1. I’ll sort this once I also backport the code optimisations already in devel.

Issue #80, opened 2023-07-14
[BUG] StandardState.py is broken 


jmichel80:
I have fixed an import error triggered by updates of the Sire Units code in StandardState.py
https://github.com/OpenBioSim/sire/blob/feature-fixstandardstatecorrection/wrapper/Tools/StandardState.py

This has allowed me to run the program from the command line, but I have encountered various other runtime errors. 
Running 
`standardstatecorrection -C ./somd.cfg -r traj000000001.dcd -s 1 -b 5 -d 0.25 -o 4 `
with 
[somd.zip](https://github.com/OpenBioSim/sire/files/12053158/somd.zip)
(dcd file not included in the attached) 

generates the following output

```
64], [30.611033320426944, 40.67983399118694, 40.89288466317314]]}
Traceback (most recent call last):
  File "/home/julien/miniforge3/envs/sire-dev/share/Sire/scripts/standardstatecorrection.py", line 137, in <module>
    StandardState.run(params)
  File "/home/julien/miniforge3/envs/sire-dev/lib/python3.10/site-packages/sire/legacy/Tools/__init__.py", line 175, in inner
    retval = func()
  File "/home/julien/miniforge3/envs/sire-dev/lib/python3.10/site-packages/sire/legacy/Tools/StandardState.py", line 361, in run
    weight_norm_factor = ROT / sum([x[1] for x in guest_orientations])
TypeError: unsupported operand type(s) for +: 'int' and 'tuple'
```

The TypeError happens at a line of code introduced in this previous commit e2e370940894315838fb8f65e141baaf07050ce0 by @fjclark 

I don't understand what would have changed in Sire since November 2022 to cause this previous code to break ?

chryswoods:
I've taken a look and don't think that this is something that has changed in sire. The `guest_orientations` object is a list of list of tuples. For example, here's the `guest_orientations` for a trial run I used with the config file you supplied;

```
GUEST = [[(( -6.414, 4.968, 1.616 ), 0.0), (( -5.984, 4.508, 0.106 ), 0.0), (( 3.606, -11.622, 1.826 ), 0.0), (( 4.616, 1.858, -1.764 ), 0.0), (( 4.176, 0.288, -1.784 ), 0.0)], [(( -4.968, -6.414, 1.616 ), 0.0), (( -4.508, -5.984, 0.106 ), 0.0), (( 11.622, 3.606, 1.826 ), 0.0), (( -1.858, 4.616, -1.764 ), 0.0),...
[...]

PR #81, opened 2023-07-16
Fix standardstatecorrection

fjclark:

The current version of `standardstatecorrection` is broken because I failed to include all required changes in the original PR (due to a mix up with personal branches) - this PR adds these changes. I have tested the updated script with an old trajectory and obtained exactly the same correction as previously (using the version [here](https://github.com/fjclark/Sire/blob/mdr_bugfixed_from_boresch/wrapper/Tools/StandardState.py)).

## If this is to fix a bug...

This pull request fixes issue https://github.com/OpenBioSim/sire/issues/80 . 

## Suggested reviewers:
@chryswoods

### DIFF
--- a/wrapper/Tools/StandardState.py
+++ b/wrapper/Tools/StandardState.py
@@ -1,14 +1,15 @@
-#Script to calculate the free energy change
+# Script to calculate the free energy change
 # upon removing a set of host-guest distance
 # restraints and applying standard state conditions
 # @authors: Stefano Bosisio and Julien Michel
 
-import os,sys, random
+import os, sys, random
 import math
 from math import pi, cos, sin
 from Sire.Tools.OpenMMMD import *
 from Sire.Tools import Parameter, resolveParameters
 from Sire.Maths import *
+
 # Python dependencies
 
 try:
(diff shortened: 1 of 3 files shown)


chryswoods:
The failure is due to an issue with the conda recipe. I’ve fixed it in the other PR currently building. I’ll apply the fix to your branch when I accept and merge the other PR (assuming it all builds ok)

fjclark:
Great, thanks for letting me know.

chryswoods:
I've fixed things in devel and have updated your branch / PR. It is now building. Hopefully it will all be green ticks tomorrow morning :-)

chryswoods (dismissed):
I'm happy for this to be merged once @jmichel80 is happy.

fjclark:
Thanks @chryswoods. I've also just added a test. This uses very few integration points and a short trajectory to get it to run in under 3 seconds - as a result the correction is huge and incorrect, but at least it ensures that the code runs.

wrapper/Tools/StandardState.py
```diff
@@ -177,265 +205,306 @@ def genOrientations(restr_dict, norientations=5):
...
+    kbT = kb * T
+    beta = 1 / kbT
 
     Ztot = 0.0
     Uavg = 0
 
+    # If norient <= 3, then there will be no orientations with any weight.
+    if norient.val <= 3:
+        print("Error, norient must be > 3. Abort.")
+        sys.ext(-1)
```
jmichel80, line 270: typo / sys.exit(-1)

wrapper/Tools/StandardState.py
```diff
@@ -177,265 +205,306 @@ def genOrientations(restr_dict, norientations=5):
...
             restr_dict[pairs][1].append(coord1)
         current_frame += step_frame
 
-    #now restr_dict has:
-    #restr_dict[lig,host]=[ [req,K,D], [ [coords]...] ,[ [coords],...] ]
+    # now restr_dict has:
+    # restr_dict[lig,host]=[ [req,K,D], [ [coords]...] ,[ [coords],...] ]
     print("Calculating average coordinates for restrained atoms")
-    #import pdb; pdb.set_trace()
+    # import pdb; pdb.set_trace()
```
jmichel80, line 393: delete from script

jmichel80:
The CI failed for python 3.11 on windows with the following message

```
..\_test_env\Lib\subprocess.py:1538: FileNotFoundError
=========================== short test summary info ===========================
ERROR tests/somd/test_standardstatecorrection.py::test_standard_state_correction - FileNotFoundError: [WinError 2] The system cannot find the file specified
```
builds on other OSs completed without error.

jmichel80:
Quick [googling](https://stackoverflow.com/questions/33560364/python-windows-parsing-command-lines-with-shlex) suggests that shlex may not be reliable on windows.  
@chryswoods would you have suggestions for a portable command line parser that could be used to rewrite this [test](https://github.com/OpenBioSim/sire/blob/devel/tests/somd/test_standardstatecorrection.py) ?

chryswoods:
I've pushed a fix for this test to `devel`. There were a few things to fix:

1. The test needed marking as being slow, so that it is skipped for quick tests.
2. shlex.split is unreliable and there is no good option on Windows the works all the time in CI. As such, we disable tests that use shlex if this is run on Windows, e.g. see tests/somd/test_somd.py.
3. The test needs to be marked to be skipped if its dependencies aren't available. These are scipy and wget. We can remove the wget dependency if the test is migrated to use the same remote download code that we use to download bzipped structures. Maybe @mb2055 could take a look and update the test to download the DCD file from https://sire.openbiosim.org/m as it does for other structure files?

chryswoods:
I've back-ported this fix as part of the #82 PR (back-port all to `main`). I've also run a new build/deploy GH Action on devel just for Windows, so that the Windows devel package will be uploaded by tomorrow morning.

Issue #85, opened 2023-07-25
analyse_freenrg with TI is incorrect when simulation times are unequal


fjclark:
**Describe the bug**

When simulations at different lambda windows are not all run for an equal amount of time, the TI result from `analyse_freenrg` is incorrect by up to several kcal / mol. This is because the arrays used to hold the gradients [are initialised with zeros](https://github.com/OpenBioSim/sire/blob/19ba5a1d25aa638f02019b530b79bc064cecdbd4/wrapper/python/scripts/analyse_freenrg.py#L759) according to the length of the longest simulation. For all simulations other than the longest, the last values will remain 0. This means that [when finding the mean](https://github.com/OpenBioSim/sire/blob/19ba5a1d25aa638f02019b530b79bc064cecdbd4/wrapper/Tools/FreeEnergyAnalysis.py#L91), the result is much closer to 0 than it should be.

I will submit a PR to fix this shortly.

**(please complete the following information):**
 - Version of sire: latest `dev` release
 - I confirm that I have checked this bug still exists in the latest released version of sire: yes

chryswoods:
I'll close this issue as the fix is now merged into `devel` and `main` (and released in 2023.3.1)

PR #86, opened 2023-07-25
Fix analyse_freenrg TI for unequal sample sizes

fjclark:
Several arrays were initialised with zeros according to the maximum sample size. When simulations at different lambda windows were run for unequal lengths of time, this meant that the end of most arrays consisted of 0s. Taking the mean of these arrays produced means which were erroneously close to 0, producing highly incorrect TI estimates.

To avoid this, I've replaced arrays initialised with 0s with arrays initialised with nans, and used the functions `nanmean` and `nanstd` to ignore the nans. I have tested on previous output to confirm that the MBAR result is identical to previously, while the TI results show dramatic changes.

This pull request fixes issue #85.

## Suggested reviewers:
@chryswoods, @lohedges

Many of the changes are due to auto-formatting with black.

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -24,6 +24,9 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 * Added a :func:`sire.minimum_distance` function to calculate the minimum
   distance between atoms of two views.
 
+* Fixed a bug in ``analyse_freenrg`` which produced incorrect TI results
+  when not all lambda windows were run for equal lengths of time. 
+
 * Please add the changelog entry for your PR here. We will add the link to your PR
   during the code review :-)
 
--- a/wrapper/Tools/FreeEnergyAnalysis.py
+++ b/wrapper/Tools/FreeEnergyAnalysis.py
@@ -34,9 +34,7 @@ except ImportError:
     import platform
 
     if platform.machine() in ["aarch64", "arm64"]:
-        print(
-            "'pymbar' is not available on the 'aarch64' or 'arm64' platforms"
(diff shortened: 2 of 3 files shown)


chryswoods:
Looks good - I think the CI failure is to do with tags I broke a while back. I will pull this into the PR for 2023.3.1 and will build and test there. This should also mean that this fix will make it for 2023.3.1 ;-)

fjclark:
Perfect, thanks!

lohedges:
>  I think the CI failure is to do with tags I broke a while back.

I think this due to some change in conda and the way that environment variables are parsed. It's a particular problem for PRs from external forks, where the tags aren't always present. I had a similar issue for BioSimSpace and needed to use the following logic in the recipe to fix things:

```
package:
  name: {{ name }}
  version: {{ environ.get('GIT_DESCRIBE_TAG', 'PR') }}

source:
  git_url: BSS_REMOTE
  git_tag: BSS_BRANCH

build:
  number: {{ environ.get('GIT_DESCRIBE_NUMBER', 0) }}
```

According to the docs (and issue threads I found) it shouldn't be needed, but seemed to be the only thing that worked for me. I was meant to mention it to you (since my original code was based on your template), but apparently forgot. Ironically, I first noticed the issue with an external PR from @fjclark to the BioSimSpace repo.

Cheers.

chryswoods:
Thanks @lohedges - I've updated the recipe in `devel` and have pulled the changes in here. Hopefully the build will work :-)

If it does, I will review and accept the PR, and will then backport to `main` via the open PR for 2023.3.1.

Issue #84, opened 2023-07-17
[BUG] Sire.IO.PDB2 not handling systems with more than 100000 atoms


cespos:
**Describe the bug**
If a system has more than 100'000 atoms, BioSimSpace.IO.saveMolecules() messes up the PDB format

**To Reproduce**
```
import BioSimSpace as BSS

ligand_coord = BSS.IO.readMolecules('aspirine.sdf')[0]
ligand = BSS.Parameters.parameterise(ligand_coord, 'gaff2').getMolecule()
box = [110 * BSS.Units.Length.angstrom] * 3
angles = [90 * BSS.Units.Angle.degree] * 3
solvated = BSS.Solvent.solvate('tip3p', ligand, box=box, angles=angles)
BSS.IO.saveMolecules("test.pdb", solvated , 'pdb')
```

**Expected behavior**
Correct PDB formatting

**Input files**
aspirine.sdf (Can be downloaded from [Pubchem](https://pubchem.ncbi.nlm.nih.gov/compound/Aspirin#section=3D-Conformer))

**(please complete the following information):**
 - Ubuntu
 - Version of Python: 3.10.12
 - Version of BioSimSpace: 2023.3.0
 - I confirm that I have checked this bug still exists in the latest released version of BioSimSpace: [yes/no] yes

**Additional context**
![image](https://github.com/OpenBioSim/biosimspace/assets/31316943/a8cbc8e0-2875-466f-ab35-bbc3b43cfd12)

lohedges:
This is caused by the renumbering of atoms when TER records are detected. The original serial is set correctly, but then the same capping isn't applied when renumbering.

lohedges:
I've fixed this on the [fix_84](https://github.com/OpenBioSim/sire/tree/fix_84) branch. I'm just checking that there aren't any unexpected side effects.

chryswoods:
Ok - I will wait for this for `2023.3.1`

PR #87, opened 2023-07-26
Fix issue #84

lohedges:
This PR closes issue #84 by making sure that atom serial numbers are capped to 99999 when renumbering when TER records are present in the PDB file. I've also fixed another small bug (which didn't seem to cause an issue) which meant that the atom serial wasn't incremented when certain TER records were written, which meant that the atom following the TER had the same serial.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/corelib/src/libs/SireIO/pdb2.cpp
+++ b/corelib/src/libs/SireIO/pdb2.cpp
@@ -2263,9 +2263,12 @@ int PDB2::parseWholeMolecule(const SireMol::Molecule &molecule, QVector<QString>
             PDBAtom atom(atoms(i), is_ter[i], map, errors);
             last_atom = atom;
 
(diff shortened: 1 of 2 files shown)


chryswoods (approved):
Looks good - nice to see you use `atom.setSerial` so that there is a single place where the atom renumbering logic is located.

Not for now, but maybe in the future we want to add a PropertyMap property to let the user control what to do in this case (e.g. reset the counter, cap at 9999, or allow overflow to 10000)?

lohedges:
Yes, I was thinking something similar too. I just went with what other parsers do, but it would be good to let the user specify what do do when the serial overflows.

chryswoods:
Closing as this has now been merged with the other PR

PR #91, opened 2023-07-28
Update analyse_freenrg to use pymbar 4

fjclark:

Since it was a quick fix, I've updated `analyse_freenrg` to use the pymbar 4 API. I tested on two sets of free energy data before and after the changes and got identical results.

## Suggested reviewers:
@chryswoods, @lohedges


chryswoods (approved):
I approve of all of these changes. Thanks for updating the requirements_bss and try_import lines to the latest versions :-)

chryswoods:
@jmichel80 - is this something you are happy to merge? Do you want to run more tests? This is a change in functionality, so would go into `2023.4.0.dev` for a development period, and would only be properly released and supported once `2023.4.0` is released at the end of September. I am aware though that this is close to the workshops...

jmichel80:
Could we support both pymbar versions easily for the time being ? The default would be pymbar 4 but it would make it possible to revert to using BSS in an environment with pymbar3 in case we run into issues.

fjclark:
> Could we support both pymbar versions easily for the time being ? The default would be pymbar 4 but it would make it possible to revert to using BSS in an environment with pymbar3 in case we run into issues.

@jmichel80 I've updated the code to be compatible with either version of pymbar. I've tested with pymbar greater than, and less than 4 on the same data (bound vanish leg of a MIF ABFE calculation) and got identical results.

fjclark:
Great, thanks very much @chryswoods. No problem at all.

PR #95, opened 2023-08-02
Ensure all python modules are loaded when selecting the new api

### LINKED ISSUE #94: [BUG] Failure to properly import when running in a parallel region
**Describe the bug**
Calling sire functions within a parallel region (for example using a `worker` within `concurrent.futures`) results in exceptions due to failures to import sire modules (despite sire being imported prior to parallelisation).

**To Reproduce**
Run the following code:
```
import sire as sr
import concurrent.futures

def get_info(mols):
    map = {
        "integrator": "langevin_middle",
        "platform": "CPU",
        "threads": 1,
        "temperature": 300 * sr.units.kelvin,
    }
    omm = sr.convert.to(mols, "openmm", map=map)
    s = omm.getState(getEnergy=True)
    return s.getPotentialEnergy()

if __name__ == "__main__":
    workers = 4
    mols = sr.load(sr.expand(sr.tutorial_url, "ala.top", "ala.crd"), silent=True)
    with concurrent.futures.ThreadPoolExecutor(max_workers=workers) as executor:
        jobs = []
        results_done = []
        for i in range(workers):
            kw = {"mols": mols}
            jobs.append(executor.submit(get_info, **kw))
        for job in concurrent.futures.as_completed(jobs):
            result = job.result()
            results_done.append(result)
```
This gives the error
`AttributeError: module 'sire.convert' has no attribute 'to'`

**Expected behavior**
This code should run, in this case giving 4 identical potential energies (this is precisely what it does when run in serial)

 - OS: Linux (ubuntu)
 - Version of Python: 3.10.12
 - Version of sire: 2023.4.0.dev

**Additional context**
This is almost certainly an issue with lazy importing, as of now the issue can be fixed by either importing `sire.convert.to` within the parallel region, or calling this functionality prior to the parallel region.
The `SIRE_NO_LAZY_IMPORT` doesn't appear to be functioning properly, so it doesn't fix the issue, instead it gives the error `AttributeError: module 'sire' has no attribute 'units'`.

chryswoods:
This fix ensures that all pythonic sire modules are loaded if `use_new_api()` is called. This allows the user to control when all modules are loaded, which is useful for parallel processing.

This should fix issue #94. The code didn't have an `else` in the lazy-load part, because we have to wait until the user has chosen either the old, mixed or new APIs before we can load modules. This choice is either explicit (calling, e.g. `sr.set_new_api()`), or implicit by calling any of the new modules (calling a function from the module triggers its loading, which triggers the new API if this has not already been set).

This fix forces all modules to eagerly load once `sr.use_new_api()` has been explicitly called.

## Suggested reviewers:
@mb2055, @lohedges

Closes #94

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -31,6 +31,9 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
   values of lambda to be performed. This is initial functionality which
   will be documented and expanded by subsequent PRs.
 
+* Forced all new-style modules to import when `sr.use_new_api()` is called.
+  This will make it easier to use sire with multiprocessing.
+
 * Please add the changelog entry for your PR here. We will add the link to your PR
   during the code review :-)
 


mb2055 (approved):
Everything looks to be running smoothly in parallel now. 
Thanks for addressing this so quickly

PR #96, opened 2023-08-03
[WIP] Allow GROMACS water molecules to be flagged as crystal waters

lohedges:
This is a small update to allow water molecules to be flagged as crystal waters when swapping the topology to GROMACS format, which is required prior to performing solvation in BioSimSpace. This change means that we can still solvate around any existing water molecules, but `gmx genion` will ignore them when choosing which water molecules to replace in order to neutralise the system (or reach a specified ion concentration). Following this, we can then update the topology of the crystal waters to the standard template, i.e. `SOL`, so that we recover a single solvent group.

These changes are related to OpenBioSim/biosimspace#112 and OpenBioSim/biosimspace#148.

* I confirm that I have added a test for any new functionality in this pull request: [y] (Will be tested via BioSimSpace.)
* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: [y] (Will be documented via BioSimSpace.)

## Suggested reviewers:
@chryswoods


lohedges:
Just to say that I'm not sure if this falls under _fix_ or _feature_. Clearly the particular use case of this update is _fixing_ unintended behaviour when solvating, but I think it will be included with a more general update to the `BioSimSpace.Solvent` code that builds of OpenBioSim/biosimspace#149 Crystal waters seem to be a recurring issue at the moment, so I want to add an example page showing how to deal with them. I'm happy with whatever you decide for this particular PR.

lohedges:
Everything works as expected in BioSimSpace. Phew!

lohedges:
Hmmm, nothing seems to work, and I get a different topology file depending how the different water molecules (crystal and normal) are arranged in my system. Perhaps we'll need to use entirely different water templates for the crystal waters. If so, it's probably easier for me to just use a different approach for handling the `genion` issue, e.g. re-running until no crystal waters were removed.

This has also made me realise that the request in OpenBioSim/biosimspace#148, i.e. not changing the existing water topology on write on solvation, is not possible with the current GroTop parser, i.e. if there are other water molecules in the system then the molecule types will always be de-duplicated. (This is what you'd want for the purposes of running anything with GROMACS.) You should always be able to reference your original waters by index anyway.

chryswoods:
(or, we could add a way of marking a water molecule as something that should not match the `mols["water"]` search term. This could be a `is_not_searchable_water` property?

lohedges:
Aha, I didn't realise that you were now using the water search functionality. Yes, it should be pretty robust. I wrote it because I needed to be able to detect and swap water topologies in BioSimSpace, and this is obviously painful in Python for any reasonable number of water molecules.

In this case the crystal waters will have a very specific formatting since this is something that is only implemented as a fix behind the scenes. It should be simple enough to combine the water search with a residue name match to extract them, e.g.

```python
xtal_waters = mols["water and rename XTL"]
```

I'll have a play with this tomorrow.

Thanks for the pointers!

lohedges:
That said, having the ability to flag something as not searchable might be quite useful, e.g. by adding a specific property to a molecule. This would help in the situation where we do want to preserve some user naming for a topology, but don't know what that naming would be, i.e. don't have an easy way to search for it.

lohedges:
Great, that works perfectly, i.e. adding excluding crystal waters from the search, then adding a separate search to define a new molecule type for them. I'll see how easy it is to add in a check for an `is_non_searchable` property, so that the user can keep custom water naming if desired.

I somehow hadn't noticed that the water searching that I wrote with `IDEngineWater` has now moved into functionality within `SireMol` itself, e.g. an `is_water` function. That's much more useful.

lohedges:
Okay, this all appears to be working. The BioSimSpace feature branch that uses these changes is [here](https://github.com/OpenBioSim/biosimspace/tree/feature_crystal_water), along with an associated [unit test](https://github.com/OpenBioSim/biosimspace/blob/feature_crystal_water/tests/Solvent/test_solvent.py).

chryswoods (approved):
Looks good - I would put this under "feature" rather than "fix" as it adds new capability that involves a change in functionality. I also want some time to do some debugging of it in `devel` just to make sure that the searches are fast, and that there isn't that will bite us (it all looks ok, but it is a new code path).

lohedges:
Thanks, I agree. While this seems to work and is entirely invisible to the user, it would be good to have some time to test it in the wild and refine as necessary.

PR #102, opened 2023-09-05
Modularise triclinic lattice rotation and reduction functions and make optional

lohedges:
This PR modularises the functions for performing the lattice rotation and reduction during the instantiation of a `TriclinicBox` and makes them optional. (The rotation is needed to make the triclinic cell comply with the constraints of certain molecular dynamics engines, i.e. vector0 aligned with x axis, vector1 in x-y plane, and vector2 with positive z component. This is normally done for simulation efficiency. The reduction is also required by certain engines, e.g. OpenMM.) This will allow the user to call the functions independently on an existing space (these can be exposed via BioSimSpace) or choose to disable/enable them on read. (By default we could choose to disable both, then add options to `BioSimSpace.IO.readMolecules` to rotate/reduce the space as desired.) We can also selectively enable/disable them on a per-parser basis, as needed.

In working on this PR I have realised that one reason for the box angles flipping is due to the need to reorder the lattice vectors by magnitude when performing the rotation. If the magnitudes are very close, but different, which is the case when read from fixed-width format (particularly when converting from angles to vectors) then the vectors will likely be reordered unnecessarily. As such, I've added a `precision` option to the `rotate` function that allows the user to specify the precision to which the vectors are sorted. Setting this to the minimum precision of the text-based input files that we support (1e-5) gives stable angles. Similarly, I've exposed a `bias` option for the `reduce` function, that allows the user to specify the bias for the rounding used during rotation. This can be used to choose a consistent direction for the rounding, helping prevent oscillation. (The existing bias seems to work well.)

All of the Sire and BioSimSpace tests still pass following this update. The box parameter creation utility functions, e.g. `truncatedOctahedron` always use a rotation/reduction by default, so a user generating initial box parameters in Sire/BioSimSpace should start from sane values, i.e. they _shouldn't_ need to further rotate/reduce on read/write to preserve things.

Something that we might want to discuss is whether we want tojust disable the rotation/reduction by default in Sire layer and deal with it in BioSimSpace, i.e. set the default values for `auto_rotate` and `auto_reduce` to `false`.

@msuruzhon: Tagging you in so that you are aware of changes. Once this is approved/merged, I'll backport to 2023.2.3. I can simply disable the rotation/reduction for the AMBER formats (RST/RST7) or could also show you what you'd need to change in BioSimSpace if you'd like to be able to do this manually, e.g. if you want more control.

## Suggested reviewers:
@chryswoods


lohedges:
Oh, and @msuruzhon, I forgot to mention that a possible reason why the binary rst format gives different behaviour for 2023.2.3 and 2023.3.0 is that for 2023.3.0 restarts are now written using restart formatting, which uses doubles for all records rather than floats. This should also be easy for me to backport. (This still doesn't explain why rst7 worked, though, since that is certainly less precise than a float.)

msuruzhon:
Thank you for this @lohedges. I think it might make sense to keep this reduction off by default, seeing that otherwise `sire.convert.to_openmm` won't always work by default otherwise. 

Could you please tag me in your relevant BSS PR that turns off this reduction for AMBER, so that 1. I can cherrypick it to our release branch and 2. see how I can tweak these settings for my OpenMM testing. Many thanks!

msuruzhon:
> Oh, and @msuruzhon, I forgot to mention that a possible reason why the binary rst format gives different behaviour for 2023.2.3 and 2023.3.0 is that for 2023.3.0 restarts are now written using restart formatting, which uses doubles for all records rather than floats. This should also be easy for me to backport. (This still doesn't explain why rst7 worked, though, since that is certainly less precise than a float.)

Thanks, I guess if we turn off the reduction we won't need that?

lohedges:
I'm just rebuilding and testing 2023.2.3 locally now. If I disable the rotation/reduction for the AMBER parsers in Sire, then you won't need to modify BioSimSpace at all. I'll still test the AMBER rst double precision fix, since it's worth enabling anyway if it works, e.g. you'd probably prefer to have coordinates and velocities in a higher precision anyway.

lohedges:
Oh, and for the OpenMM bit, I'll see if manually doing the reduction prior to conversion (without the rotation) fixes things.

chryswoods:
Thanks @lohedges - this looks great. It is good to have the different stages broken out and individually toggleable.

I think you are right - we probably shouldn't do the automatic reduction in sire because it leads to these unexpected errors when loading / saving files. It is against the principle of trying to write the same information that we read.

I think we should disable auto rotation / reduction for all of the parsers and then see if that breaks anything. I think it will be easier to fix anything that breaks by adding manual rotation / reduction at the points where it is needed.

I am happy to merge this once the checks pass and you and @msuruzhon are happy.

chryswoods (dismissed):
👍

lohedges:
Thanks. I've realised that I've forgotten to call `this->setAttributes` within the `rotate` method so that things are set correctly if the user rotates but doesn't reduce (unlikely, but possible). I'll update tomorrow and let you know when I've checked everything.

lohedges:
Okay, it looks like setting both `auto_rotate=false` and `auto_reduce=false` is safe. This is because all of the test input uses correctly reduced box vectors to start with and doesn't convert between formats when running simulations. Once Exs have tested and are happy with this, I'll add a feature branch to BioSImSpace to expose the `rotate` and `reduce` methods and add options to `BioSimSpace.IO.readMolecules` to allow a user to choose to have these called on read.

msuruzhon:
Hi @lohedges, I have just tested this code internally and so far there are no problems with our AMBER simulations! Many thanks!

chryswoods (approved):
👍

lohedges:
That's great, thanks for the update. I'm just running a few more checks locally and will also see if I can figure out why OpenMM is still unhappy with the reduction in certain cases.

lohedges:
I've decided to default to a full precision rotation and an unbiased reduction. The latter means that our reduction is fully compatible with OpenMM, so can be used within `sr.convert` when OpenMM complains that box vectors aren't reduced. I know what parameters allow us to have a stable rotation/reduction if needed (these are used in the existing tests), so a user can set these if they want.

On the BioSimSpace side, this means that the default box angles generated by `BioSimSpace.Box` are different for certain triclinic spaces. I'll need to update the tests that validate the `setBox` does indeed set the correct box parameters.

lohedges:
Everything is now off for all parsers and we can deal with any rotation/reduction issues on a per-parser basis as needed within BioSimSpace. When I mentioned OpenMM above, I was just saying that our `space.reduce()` function will give a reduction that will work with OpenMM, so can be used in the Sire conversion layer. (We are now no longer automatically rotating/reducing anything.)

chryswoods (approved):
Looks good - I'm happy for you to merge this :-)

PR #110, opened 2023-10-02
Feature Permanent Distance Restraints

fjclark:

This PR adds support for one or more "permanent" distance restraints. These are distance restraints that are always applied, and are never scaled by λ. This allows the release of all other distance restraints to a single harmonic or flat-bottomed restraint. When the ligand is fully decoupled, the free energy of release of the single remaining restraint can be computed without simulation. This provides more options for using multiple distance restraints in ABFE calculations.

I will create a parallel PR to BioSimSpace with functionality for setting up and running ABFE calculations in this way.

- Introduce "permanent distance restraints" which are never scaled by λ.
- Add new 'always dummy' group for atoms which begin and end as dummies

* I confirm that I have added a test for any new functionality in this pull request: n
* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: n

I haven't yet added new tests as I was unsure how best to do this. Ideally, I would add some tests similar to [this](https://github.com/OpenBioSim/sire_bigtests/blob/main/SireMove/test_recept_lig_restraints.py). Would a test like this be too slow? What would you recommend?

## Suggested reviewers:
@chryswoods, @lohedges

I'm currently testing SOMD ABFE calculations with Boresch restraints against multiple distance restraints as shown [here](https://github.com/fjclark/bss_multiple_distance_restraints_example). Both methods produce results which are statistically indistinguishable with the TYK2 system tested:

image

Thanks very much.


chryswoods (approved):
👍

PR #111, opened 2023-10-07
Feature bond breaking

chryswoods:

This PR completes the work undertaken to add alchemical free energy simulation support to sire. The new functionality is described in the changelog, but mostly it is fully described in the new Part 6 of the tutorial. Tests for new functionality have been added. This includes an end-point energy test for the pentane to cyclopentane perturbation.

## Suggested reviewers:
@lohedges

This includes the recent PRs merged into devel, so I expect that this will also fail CI on Windows. I will debug locally and will push a fix.

I will also create a new branch from this as a 2023.4.0 prep, just making documentation fixes to the website (e.g. updated features, Quickstart guide and typo corrections). I am not going to make any functional changes now before 2023.4.0.


chryswoods:
I think I have fixed the compile error on Windows...

chryswoods:
That fixed that error on Windows, but now there is a strange multiply-defined symbol link error for the boost parser in SireUnits... (I can't reproduce this locally yet, as my local windows build completes without error).

chryswoods:
I can reproduce the Windows compile errors locally (thankfully). It is a Python 3.11, using a new version of boost I think? There's lots of small boost compile errors that look like they may be incompatibilities between boost and the version of MSVC used to compile the conda package. I'll see if the fixes I've made will let it compile overnight...

chryswoods:
Ok - I've fixed the Windows compile issue. There was one boost header that I couldn't fix (mcgregor_common_subgraphs). This is only used in `findMCS.cpp`. I've thus disable this functionality on Windows (sire's MCS matching). I don't think it is used much, and people prefer RDKit. It is a shame to lose the fallback, but fixing this would, I think, require upgrading to a new MSVC compiler version for the Windows build. This is too much work for this release. I may consider it for the point release.

The build failure is just a couple of unit tests failing on Windows that require the findMCS functionality. I've conditionally disabled these tests. I haven't re-run the CI to save GH Actions, and because I have tested locally that everything is ok.

I may look into automatically detecting the failure of mcgregor_common_subgraphs in cmake, as it may be that this works with Python 3.9 or 3.10 (my Windows build used to be 3.10, which is why I couldn't reproduce it locally - but that could also have been that my build didn't upgrade boost). 

I don't want to pin to an older version of boost on Windows as this will be a compatibility issue with other packages (boost is a very common dependency).

If the functionality is really needed on Windows, then another option may be to copy in the older mcgregor_common_subgraphs header file and patch it to work with new boost, without pulling in boost/describe.hpp. But, that may be quite a rabbit hole... ;-)

I've run sire_bigtests and they all pass on all OSs. I'm also now looking at the BioSimSpace tests from devel and the string units PR. There's a few things that fail because of `new_system._sire_object.property("space").isPeriodic()` fails because the space is a `Cartesian()`, which I will look into tomorrow. I think we are nearly there :-)

lohedges:
The Sire MCS fallback is only really triggered when the user passes in a `prematch`, since RDKit has no easy way to build the MCS and checking against it at each step, i.e. you only check that the MCS contains the pre-match as a post-processing step. We can just disable this on Windows for now and re-enable once a fix is found.

The Cartesian space issue sounds strange. I recently needed to update the way that I copy new boxes from trajectory and restart files into the reference system, since I don't want to override a valid space if there isn't one in the file, where Sire would apply a default Cartesian space for you. The logic seemed to work and tested passed during the PR. I guess something else must have changed.

Glad we're getting there!

chryswoods:
Sorry - a false alarm on the BioSimSpace tests. I'd mistakenly not uninstalled an older version of BioSimSpace before installing `feature_string_protocol`. I can confirm that, with `feature_string_protocol` installed, all of the tests that I can run pass (I haven't got ambertools installed so can't do anything with OpenFF). The logic for spaces all works, as does the new code that lets you pass in strings for units :-)

lohedges (approved):
I've tested locally and can confirm that all BioSimSpace tests pass. Thanks for this Herculean effort!

Issue #116, opened 2023-10-19
[BUG] Issue merging small molecules


noahharrison64:
Hi,

I've been encountering issues when trying to use the Align.merge() functionality. 

Code:
```py
paramd = []
for filebase in ['ejm31', 'ejm42']:
    mol = BSS.IO.readMolecules([f"{filebase}.sdf"])
    param_mol = BSS.Parameters.gaff2(mol.getMolecule(0)).getMolecule()
    paramd.append(param_mol)
    
ejm31, ejm42 = paramd
merged = BSS.Align.merge(
    ejm31,
    ejm42,
    force = True
)
```
Error:
Large stack trace so added as [text file](https://drive.google.com/file/d/1xLa-1KnOkONDnTRYO6wxgF10aH--mYDw/view?usp=share_link)

**Input files**
I am using the tyk2 data files provided in the old BioSimSpaceTutorials repo.
[ejm31.sdf](https://github.com/michellab/BioSimSpaceTutorials/blob/main/04_fep/02_RBFE/inputs/ligands/ejm31.sdf), [ejm42.sdf](https://github.com/michellab/BioSimSpaceTutorials/blob/main/04_fep/02_RBFE/inputs/ligands/ejm42.sdf)

**(please complete the following information):**
 - OS: Linux x84_64
 - Version of Python: 3.9.1
 - Version of BioSimSpace: 2023.4.0
 - I confirm that I have checked this bug still exists in the latest released version of BioSimSpace: yes

Many thanks,
Noah

chryswoods:
Thanks for reporting. I think from the stack trace that something is going wrong with the transfer of the RDKit shared pointer from C++ up to Python. This is in the sire layer, so I've transferred the issue across to this repo so I can debug and track this here.

The wrapping of the RDKit shared pointer is, I think, handled by RDKit. Could you let me know what version of RDKit you have installed?

Do this by running;

```
>>> import rdkit
>>> print(rdkit.__version__)
```

I get `2023.03.3` on my computer, and your test script runs without issue. My guess is that you have a different version?

We're going to release `2023.4.1` soon, so I will get a fix in for that that will solve the problem :-)

lohedges:
Just to say that I've also tested locally on Linux (Python 3.11) and it's working for me too. I have:

```python
In [1]: import rdkit

In [2]: rdkit.__version__
Out[2]: '2023.09.1'
```

chryswoods:
[...]

PR #114, opened 2023-10-17
OpenMM optimisations and bug fixes

chryswoods:
This is collection of optimisations for the OpenMM code, plus some bug fixes and regressions as described in the changelog (copied below).

```
* Fixed regression introduced in 2023.4.0 that meant that removed the constraints
  from water molecules that had no internal bonds. These waters would blow up
  as there was nothing holding them together. The need for these constraints is
  now better detected and explicitly added.

* Significantly sped up the OpenMM layer by checking for similar constraint lengths
  and matching them all to be the same (within 0.05 A for calculated constraints,
  e.g. unbonded atoms or angle constraints) or to R0 for bonds where the bond
  length is within 0.1 A of R0 and the molecule isn't perturbable.

* Added a custom minimiser that is based on OpenMM's LocalEnergyMinimizer,
  but that copes better with exclusion errors, and that has deep integration
  with the progress bar / interuption system.

* Fixed a bug where the exclusions and exceptions were mismatched for the
  OpenMM CPU platform, leading to exclusion errors.
```

I've added new tests that check for the above and show the issues have been fixed.

* I confirm that I have merged the latest version of `devel` into this branch before issuing this pull request (e.g. by running `git pull origin devel`): [y]
* I confirm that I have added a changelog entry to the changelog (we will add a link to this PR as part of the review): [y]
* I confirm that I have permission to release this code under the GPL3 license: [y]

## Suggested reviewers:
@lohedges, @mb2055 

## Any additional context of information?

Once accepted, I am going to backport this to `main` so that it can be in `2023.4.1`

PR #117, opened 2023-10-20
Fix 116

chryswoods:
This pull request fixes issue #116 and also fixes the vacuum bug reported on slack.

## Suggested reviewers:
@lohedges

This builds on and supersedes PR #114

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,4 @@ tests/cache/*
 
 .DS_Store
 
+.coverage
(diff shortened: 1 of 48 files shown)


lohedges (approved):
All BioSimSpace tests are passing and the somd2 minimisation issue is resolved. The one thing to note is that `tests/convert/test_openmm_minimise.py` appears to pass both before and after the fix. I'm not sure why this system isn't affected by the issue that I was seeing. Perhaps it's worth updating the test with the perturbable molecule example so that we know we're using something that used to fail.

Issue #106, opened 2023-09-20
[BUG] AmberRst7.box_angles() objects have the wrong units


lohedges:
Angles returned by the `box_angles()` method have length units:

```python
In [1]: import sire as sr

In [2]: s = sr.load_test_files("ala.crd", "ala.top")

In [3]: rst7 = sr.legacy.IO.AmberRst7(s._system)

In [4]: rst7.box_angles()
Out[4]: ( 90 Å, 90 Å, 90 Å )

In [5]: rst7.box_angles()[0].angle()
Out[5]: 0

In [6]: rst7.box_angles()[0].length()
Out[6]: 1
```

chryswoods:
Good spot - this is in the python layer only, but is definitely annoying. The `Vector` class has some Python overrides that add length units to its components. I'll need to either remove these for this class or create a typed-vector interface for angles. For now, use `.value()` to get the raw doubles representing the angles in degrees.

lohedges:
Great. I assumed it must be a Python layer thing since I've never had an issue with parsing anything, or creating spaces based on the box information. As suggested, I'm currently using `.value()`, then manually converting to degrees :+1:

PR #119, opened 2023-10-25
Feature 106

chryswoods:
This pull request fixes issue #106

## Suggested reviewers:
@lohedges

I've skipped CI as you will need to check this with BioSimSpace first to see if this is a workable solution. This returns a list of angle objects, so there is some internal conversion from the raw double values in degrees to double values in radians. Internally in AmberRst7, the data is always stored in raw double values in degrees. The conversion is only happening for the list of angles that we return when calling `boxAngles`.

As this is a list, it has a different API to a Vector, so it may break existing BioSimSpace code.

[ci skip]


lohedges:
Thanks. I'm only using this in the `Trajectory` code at present. I'll just need to add in conversion from radians to degrees.

Issue #103, opened 2023-09-07
[BUG] cursor.rotate should rotate all vector properties


lohedges:
Currently the [cursor.rotate](https://github.com/OpenBioSim/sire/blob/a3b5f8f68a1195fedc9b6ddf8156628c063a1cae/src/sire/mol/_cursor.py#L1533) function only rotates the `coordinates` property, when it should really rotate _all_ vector properties, i.e. `velocity` too. At present this can be achieved by calling the function multiple times, using the `map` option to remap coordinates to velocities, e.g.:

```python
# Here m is a rotation matrix.

# First the coordinates.
cursor = cursor.rotate(matrix=m)

# Now the velocities.
cursor = cursor.rotate(matrix=m, map={"coordinates": "velocity"})
```

lohedges:
Actually, the above approach doesn't work because, unless the `center` keyword argument is used, the `coordinates` property is used to compute the center-of-mass to use for the rotation, e.g. [here](https://github.com/OpenBioSim/sire/blob/a3b5f8f68a1195fedc9b6ddf8156628c063a1cae/src/sire/mol/_cursor.py#L1592-L1595).

lohedges:
This is relevant to [this](https://github.com/OpenBioSim/biosimspace/compare/devel...feature_reduce) feature branch that I'm working on, which naively uses the workaround in the original post. For now I can just compute the center of mass before calling the function and pass it in.

lohedges:
Which also makes me realise that I'll need to be careful to choose the correct center when applying the rotation, since this won't be (necessarily) the same as what is used for the box vectors themselves which would be rotated about (0, 0, 0).

lohedges:
I _think_ I've now got this working (using a workaround) on my [feature_reduce](https://github.com/OpenBioSim/biosimspace/compare/devel...feature_reduce) branch. I've also added a [unit test](https://github.com/OpenBioSim/biosimspace/commit/38a88e0a6193456edac28c9bfdc8c375693fb0c7) to check that the rotation is performed correctly and that the coordinates property is updated.

The two things that I'm not sure of are:

* For the vector properties, I aim performing the rotation about the box _origin_, which can be specified by the user (defaulting to `(0,0,0)`). It would be nice to be able to determine this automatically.
* Following the rotation of the box, the z component of the third lattice vector needs to be flipped if it is negative. I'm not sure how this should impact the properties.

chryswoods:
[...]

PR #120, opened 2023-10-25
Feature 103

chryswoods:
This pull request fixes issue #103

## Suggested reviewers:
@lohedges

This adds in rotation of Velocities when rotations are performed via a cursor. This builds on the other two PRs so will accept all three once accepted. You can toggle rotation of velocities using the `rotate_velocities` flag. This defaults to `None`, which will be `True` if the user hasn't specified this in the property map (a `rotate_velocities` flag is set which is passed down to the rotation functions in `moverbase.cpp`)


lohedges:
I guess I'd need to call this twice for perturbable molecules, i.e. to rotate both end states. This isn't a problem, since it's what I need to do now anyway. (Just remapping the coordinates property as required.)

PR #123, opened 2023-11-07
Fix issue #122

### LINKED ISSUE #122: [BUG] Invalid use of QString::compare when swapping water topology
`QString::compare` is used to compare property keys when swapping water topologies, i.e. to make sure that properties aren't updated if the molecule is flagged as `non_searchable`. However, the return value is being treated as a `bool`, when it is actually an `int`, where a value of zero indicates different strings.

lohedges:
This PR closes #122 by fixing the use of `QString::compare` when comparing molecular properties. We added a new (as yet unused) `is_non_searchable_water` property to allow a user to preserve water topologies, e.g. by excluding them from a search. However, I don't want to keep this property when swapping topology, so don't copy it to the new water template. I was incorrectly using `QString::compare` since I hadn't realised that it returned an `int`, not a `bool`. The return value is zero only when the strings match.

This will be tested via BioSimSpace.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/corelib/src/libs/SireIO/biosimspace.cpp
+++ b/corelib/src/libs/SireIO/biosimspace.cpp
@@ -271,7 +271,7 @@ namespace SireIO
         // Copy across all properties that are unique to the original molecule.
         for (const auto &prop : molecule.propertyKeys())
         {
-            if (not water.hasProperty(prop) and not prop.compare("is_non_searchable_water"))
+            if (not water.hasProperty(prop) and prop.compare("is_non_searchable_water") != 0)
             {
                 edit_mol = edit_mol.setProperty(prop, molecule.property(prop)).molecule();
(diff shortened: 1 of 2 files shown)


chryswoods (approved):
That is a surprising API - I would have expected compare to return a bool too.

xiki-tempula:
@chryswoods Do you mind back port this into the main as well, please? Thank you.

chryswoods:
Yes - although I will delegate ;-)

@lohedges - do you have time to do this tomorrow? I'm at a meeting in London all day.

lohedges:
Yes, no problem.

chryswoods:
Thanks :-)

lohedges:
This is now backported into `main`.

PR #124, opened 2023-11-09
Backport fix from PR #123

lohedges:
This PR backports the fix from #123 into `main`.

* I confirm that I have merged the latest version of `main` into this branch before issuing this pull request (e.g. by running `git pull origin main`): [y]

### DIFF
--- a/corelib/src/libs/SireIO/biosimspace.cpp
+++ b/corelib/src/libs/SireIO/biosimspace.cpp
@@ -271,7 +271,7 @@ namespace SireIO
         // Copy across all properties that are unique to the original molecule.
         for (const auto &prop : molecule.propertyKeys())
         {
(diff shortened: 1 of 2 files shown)


lohedges:
Not sure if I also need to add the entry to the CHANGELOG? This is done on `devel` under the 2023.5.0 section. Perhaps you look at fixes from that section when creating a patch release and add them to the respective 2023.4.X section.

chryswoods:
Could you create a `2023.4.2` section in the changelog and add it to there? I then compare the devel changelog to the one in main when I do the next major release, and will remove anything that is duplicated from the previous patch release.

lohedges:
Yes, no problem. I just wasn't completely sure of the system since you could end up adding a section for a patch release that never happens. I guess it all gets sorted out at release time anyway, so it's good to have something to cross reference.

lohedges:
Done :+1:

PR #129, opened 2023-11-27
Fix #128

### LINKED ISSUE #128: [BUG] TriclinicBox loses rotated and / or reduced lattice vectors on streaming to file and back.
Previously we always automatically applied a lattice rotation and reduction when a `TriclinicBox` object was instantiated. As such, the binary data stream only needed to store the _original_ box vectors that were passed in and everything could be reconstructed from those. (This allowed us to store the minimum information.) Since we no longer automatically rotate and reduce, `TricnlicBox` objects are no longer reconstructed correctly if they were rotated or reduced before being streamed to file. We now need to store all instance attributes, or some flags to say that a rotation and/or reduction have been performed.

lohedges:
This PR closes #128 by fixing the streaming operators for the `sire.vol.TriclinicBox` class. Since rotation and reduction operations are no longer automatically applied we need to add _all_ class member data to the binary data stream, not the minimal information from which the box can be recovered.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/corelib/src/libs/SireVol/triclinicbox.cpp
+++ b/corelib/src/libs/SireVol/triclinicbox.cpp
@@ -51,7 +51,7 @@ static const RegisterMetaType<TriclinicBox> r_box;
 /** Serialise to a binary datastream */
 QDataStream &operator<<(QDataStream &ds, const TriclinicBox &box)
 {
-    writeHeader(ds, r_box, 1) << box.v0 << box.v1 << box.v2
+    writeHeader(ds, r_box, 2) << box.v0 << box.v1 << box.v2
                               << box.rotation_matrix << box.cell_matrix << box.cell_matrix_inverse
                               << box.dist_max << box._alpha << box._beta << box._gamma << box.vol
                               << box.is_rotated << box.is_reduced << box.invlength;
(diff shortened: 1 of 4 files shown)


chryswoods (changes requested):
I see the need for this, but we have to be careful changing the binary streaming format. This change changes the format (as it adds more things to be streamed for this object). We need to (1) change the stream version number from 1 to 2 to indicate the change in format and (2) preserve the old format on read so that we can still read old s3 files (e.g. read assuming that this hasn't been rotated)

lohedges:
Ah yes, good point. I'll also need to auto rotate and reduce if loading a v1 stream.

lohedges:
I've updated the version number and added support for loading versions 1 and 2. Weirdly, the system in [test_triclinic.py](https://github.com/OpenBioSim/sire/blob/4a1025e8de9b5fb97c2670fa1de68b6bbf3c9d9f/tests/vol/test_triclinic.py#L95-L104) _doesn't_ fail if I update the version number and _don't_ add support for the old version. This stream file was created quite a while ago so should surely have `v1` in the header :man_shrugging:

chryswoods (approved):
Strange that this still streamed the old s3 file without the fix.

This all looks good :-)

PR #131, opened 2023-11-27
Fix issue #130

### LINKED ISSUE #130: [BUG] perturbable_constraint validation doesn't check for underscores
Unlike for the `constraint` option, the validation of `perturbable_constraint` doesn't allow underscores, e.g. https://github.com/OpenBioSim/sire/blob/4a1025e8de9b5fb97c2670fa1de68b6bbf3c9d9f/wrapper/Convert/SireOpenMM/openmmmolecule.cpp#L86-L162

This option is now parsed via `sire.options.PerturbableConstraint`, which will sanitize all input to use underscores. As such, the creation of a dynamics object will fail for _any_ perturbable constraint other than `"none"` or `"auto"`.

lohedges:
This PR closes #130 by allowing the `perturbable_constraint` dynamics option to contain hyphens.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -48,6 +48,9 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
   to keep existing behaviour, but we would recommend setting this to ``True``
   and would like to change the default in the future.
 
+* Fix validation of ``perturbable_constraint`` dynamics option when the string
+  includes hyphens. This fixes issue #130.
+
 * Please add an item to this changelog when you create your PR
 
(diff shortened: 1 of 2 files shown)


chryswoods (approved):
Looks good. I think this highlights that the way I did this for constraint was error prone. I think it should be that there is a replace("-", "_") after the simplify, so that we can match any mix of hyphens or underscores.

lohedges:
You do sanitize in the Python layer, i.e. everything gets converted to underscores. Are there situations where the C++ is being called directly?

chryswoods:
Not currently, but it is better to be safe than sorry. There would be an edge case where someone sets this as a map parameter that bypasses the sanitation.

chryswoods (approved):
All good

PR #135, opened 2023-12-05
Feature PDBx/Gemmi

chryswoods:
This PR merges in the Gemmi conversion code. This allows interconversion between Gemmi and sire objects. This means we can now read and write PDBx/mmCIF files, including loading and saving custom metadata. The load function has been updated to use PDBx/mmCIF files by default is Gemmi is available, and support for new-style PDB codes is included.

## Suggested reviewers:
@lohedges

This PR includes the changes of `feature_optimise_lever` so will minimise CI when we accept everything.

This depends on the new Gemmi shared library package being created and uploaded to conda. That package isn't up yet, so some of the CI builds here will fail. You can test locally by building the gemmi conda package locally (the recipe builds the shared-library package). Ideally I'd like to get this code into the `2023.5.0` release as it is self-contained and very useful. If the gemmi package isn't available then we may end up disabling it on unsupported platforms (MacOS, I think) and then enable it for all platforms in `2023.5.1`. Like OpenMM, Gemmi is only a runtime dependency - it only needs to be installed if the user wants PDBx/mmCIF support.

(the LJ 12/6/4 code that is mentioned in the first commit was removed in later commit - it is not included at all in this PR)


chryswoods:
Windows build failure is because the current Gemmi conda package has a static library that is built with whole program optimisation. This makes it binary incompatible with VS 2017. See this stackoverflow post:

https://stackoverflow.com/questions/54664134/cannot-compile-unknown-flag-pdbrpc-in-p2

I’ll add a cmake compile Windows that checks for this and disables Gemmi until the dynamic library package is out (similar to how this is disabled in Linux and Mac)

PR #136, opened 2023-12-08
Backporting in all of the fixes that were in devel but not in main.

chryswoods:
This should now be ready for a 2023.4.2 release that can accompany the 2023.5.0 release

## Suggested reviewers:
@lohedges


lohedges (approved):
All good. Thanks for doing this.

PR #140, opened 2024-01-06
Feature exception

chryswoods:
This PR adds support for reading and writing LJ 12-6-4 parameters from Amber topology files.                                                                                                                                                                                                                                                                                                                                                                                                                           This includes supporting Amber topology files which use LJ exceptions (i.e. the LJ parameter for an atom pair is not equal to a combining-rule combination of the atom-based parameters).

In doing this I have also detected when the topology file uses geometric combining rules and save this into the forcefield.

* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: [n/a - this is something implicit in loading/saving]

## Suggested reviewers:
@lohedges

The LJ exceptions can't be written to anything other than an Amber topology file, because only that format supports them. Also, they are not yet copied into the OpenMM context - I will do that via another feature branch. It is complex, as it will have to be via a custom exception force.


lohedges:
Thanks for this.

> The LJ exceptions can't be written to anything other than an Amber topology file, because only that format supports them.

Just to check... I seem to be able to write your test system to `GroTop` format without issue, i.e.:

```python
In [1]: import sire as sr

In [2]: s = sr.load_test_files("apo_1264.prm7", "apo_1264.rst7")

In [3]: sr.save(s, "test.grotop")
Out[3]: ['/home/lester/Code/openbiosim/sire/test.grotop']
```

Should this trigger an exception, or are there simply additional terms that are not written to the GROMACS topology file? If the latter, should we think about adding a warning to say that these terms aren't supported and will be lost so that the user isn't surprised?

PR #145, opened 2024-01-18
Fix issue #144

### LINKED ISSUE #144: [BUG] ResIdxAtomCoordMatcher segfaults if residues contain different numbers of atoms
The `ResIdxAtomCoordMatcher` is used to map atoms from the output of a `tLEaP` parameterisation to the topology of the original PDB file, e.g. potentially going from a multi-molecule representation back to a single one. There are some assumptions in this mapping, most importantly that the residues are in the same order between representations, which is the typical case of `tLEaP` for a correctly formatted PDB. However, if the PDB is invalid, then `tLEaP` might move atoms, causing the code to segfault when comparing residues since it is assuming that they have the same number of atoms.

This code could be replaced by a search over all residues, although this might be prohibitively slow for large systems. (Would speed up as residues get matched, though.)

lohedges:
This PR closes #144. The functionality here is specific to `tLEaP` so could be generalised for much greater flexibility. This will stop the segfault, at least. This has been tested via BioSimSpace.

## Suggested reviewers:
@chryswoods


chryswoods (approved):
This looks good. It is a long time since I looked at this matcher!

Issue #147, opened 2024-01-21
Sire fails to compile with python 3.12


fjclark:
**Describe the bug**
I am unable to compile the latest devel version of sire. The final section of the error message is unhelpful:
```
cd /export/users/finlayclark/software/sire-openbiosim/build/sire-boresch_sire-boresch_wrapper/Mol && /export/users/finlayclark/miniconda3/envs/sire-boresch/bin/cmake -E cmake_symlink_library _Mol.so.2024.1.0 _Mol.so.2024 _Mol.so                                                                                                  gmake[2]: Leaving directory '/export/users/finlayclark/software/sire-openbiosim/build/sire-boresch_sire-boresch_wrapper'                                           [ 90%] Built target Mol                                                                                                                                            gmake[1]: Leaving directory '/export/users/finlayclark/software/sire-openbiosim/build/sire-boresch_sire-boresch_wrapper'                                           gmake: *** [Makefile:136: all] Error 2                                                                                                                             SOMETHING WENT WRONG WHEN COMPILING THE WRAPPERS! 
```
and I am unable to find any error messages further up in the output.

**To Reproduce**
I have followed the steps given on github to compile.

**Cause**

python<3.12 was not specified in the requirements files, so the initial python 3.11 was being updated to 3.12 when `python setup.py install` was run.

**(please complete the following information):**
 - OS: Ubuntu 20.04
 - Version of Python: Intially 3.11, unintentionally updated to 3.12
 - Version of sire:  latest `dev` release
 - I confirm that I have checked this bug still exists in the latest released version of sire: yes

chryswoods:
Thanks - could you install adding the '--install-bss-deps' option? This will bring in BioSimSpace's dependencies, which aren't all available for 3.12 at the moment (although I haven't checked this for a couple of weeks).

I'll look into the compiler error - this suggests that the base sire dependencies are available for 3.12 and there is something in sire that needs fixing, so that we are ready for 3.12 once BioSimSpace's dependencies are available.

fjclark:
Thanks. I can confirm that adding '--install-bss-deps' avoids python being updated to 3.12 and allows sire to compile.

Thanks very much. Here's a more readable version of the end of the error:
```
[ 86%] Linking CXX shared library _Mol.so
[...]

PR #148, opened 2024-01-21
Specify python<3.12 in requirements files

fjclark:
This pull request fixes issue #147 by specifying python<3.12 in the requirements files. I've checked that I can now compile following the instructions in the README.

Sorry, I forgot to specify "ci skip" in my last commit.

## Suggested reviewers:
@chryswoods, @lohedges

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -15,7 +15,7 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 `2024.1.0 <https://github.com/openbiosim/sire/compare/2023.5.0...2024.1.0>`__ - March 2024
 ------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


chryswoods:
Hi Finlay - we deliberately don't add the Python version as this gets in the way of us testing newer Python versions, and the compile instructions state the supported versions.

Sire itself can compile and work well with 3.12. The challenge comes from missing dependencies. It is their requirements, rather than sire's Python requirement that stops a 3.12 installation from working. As such, we don't version Python, but instead version the other dependencies (which all are > dependencies, rather than <, so that each CI/CD run will use the latest possible version, so that any incompatibility issues will be detected quickly by use during development)

fjclark:
Hi Christopher, thanks for the explanation.

PR #143, opened 2024-01-18
Fix issue #142

lohedges:
This PR closes #142. I've killed CI since it's not necessary. Feel free to merge into one of your other feature branches and test there.

* I confirm that I have merged the latest version of `devel` into this branch before issuing this pull request (e.g. by running `git pull origin devel`): [y/
* I confirm that I have added a changelog entry to the changelog (we will add a link to this PR as part of the review): [y/n]
* I confirm that I have permission to release this code under the GPL3 license: [y]

## Suggested reviewers:
@chryswoods

PR #149, opened 2024-01-24
Add functionality to match two selections by atom coordinates

lohedges:
This PR adds a simple `AtomCoordMatcher` to match two selections by coordinates only. This is useful for mapping parameterised molecules from `tLEaP` back to the original representation of the topology, e.g. from a PDB, where the atom ordering has changed. In particular, this covers cases where our existing `ResIdxAtomCoordMatcher` fails due to the movement of an atom (in this case an ion) between residues due to the bonding.

The code makes sure to use the smaller selection as the reference and inverts the resulting mapping if the selections are swapped. There is also support for shifting the center of mass of the two selections to zero. Previous debugging found that certain molecules were shifted after parameterisation. (I cant seem to find an example of this, though.)

I've added a simple sanity check to the unit tests, but this will be tested more thoroughly by the [feature_atomcoordmatcher](https://github.com/OpenBioSim/biosimspace/tree/feature_atomcoordmatcher) branch of BioSImSpace.

## Suggested reviewers:
@chryswoods


chryswoods (dismissed):
Thanks - this is really useful :-)

I'll think about how we could expose this via the search functionality too. It could be useful to have something like

`overlapping = mols["atoms with same coordinates as {something}"]`

or even better

`overlapping = mols["(not {something}) and (atoms with same coordinates as {something})"]`

Feel free to merge this with the other approved PRs and bring them all into devel in one go

(once you have added a CHANGELOG entry)

lohedges:
Ah yes, I forgot about the CHANGELOG. Thanks for the reminder 🙂

lohedges:
I've now applied the updates from the outstanding approved feature and fix branches so this also:

closes #140
closes #143
closes #145

There were some issues resolving conflicts with the wrappers in `wrappers/Mol` since these were updated by the `feature_exception` branch as well as this one. The easiest solution was to first take all the wrappers from `feature_exception`, then rebuild based on the corelib edits in `feature_atomcoordmatcher`. As a check, I've triggered the CI here to make sure that something didn't slip through.

lohedges:
Okay, that worked. I'll merge into `devel` then backport the fixes to `main`.

PR #154, opened 2024-01-30
Preserve existing atom names when writing to PDB

lohedges:
This PR ensures that existing atom names are preserved when writing to PDB format. Previously they were always converted to upper case, which caused issues with the parameterisation of free ions where the PDB atom name is used by `LEaP` to match a template.

* I confirm that I have added a changelog entry to the changelog (we will add a link to this PR as part of the review): [y

## Suggested reviewers:
@chryswoods


chryswoods (approved):
👍

PR #160, opened 2024-02-09
Fix issue #159

### LINKED ISSUE #159: [BUG] Reduced triclinic boxes can't be used with OpenMM C++ API
As reported [here](https://github.com/OpenBioSim/biosimspace/issues/238#issuecomment-1936172960), it's currently not possible to use certain reduced-form triclinic boxes with OpenMM via the C++ API since the checks for reduced form assert that certain vector components are _exactly_ zero. This can be resolved by manually setting the components to zero if the are sufficiently close, which is actually what OpenMM does in the Python layer, i.e. before the C++ functions are called.

lohedges:
This PR closes #159 by ensuring that vector components that are sufficiently close to zero following a reduction are set to _exactly_ zero. This solves the issue of conversion to OpenMM, which requires that the first vector lies _exactly_ on the x-axis and the second is _exactly_ in the x-y plane. Non-zero values generated by our code seem to be on the order of 1e-16, so we could probably make the tolerance smaller if needed. (The OpenMM Python API uses 1e-6 as the small value.)

An alternative approach to this would be to leave the `TrinclinicBox` class untouched and simply zero the values when setting the box vectors in the Sire to OpenMM code. Let me know if you'd prefer that approach, since this issue seems to be (at present) OpenMM specific due to the way they check for reduced form.

I've cancelled the CI since it's been tested locally via some problematic `somd1` inputs. Feel free to consume into one of the other PRs if you're happy.


chryswoods (approved):
Looks good - I'll merge all the PRs together with the addition of 3.10 to the Mac build.

PR #161, opened 2024-02-10
Fix for force_lever output issues

chryswoods:
This fixes some bugs that were introduced with `feature_force_lever` that related to how the lever was printed to the screen. Despite the lever working as expected (hence why tests passed), the lever printed to the screen showing that it had no set levers.

Also, in debugging this, I found an issue in the new functionality where a default lever for a force was not being picked up (i.e. `ghost/ghost::*` was not taking precedence over, e.g. `*::charge`)

I've updated the unit tests to catch these edge cases

* I confirm that I have added a changelog entry to the changelog (we will add a link to this PR as part of the review): [N/A]

## Suggested reviewers:
@lohedges

This is a minor change to `LambdaSchedule`. I've not updated the changelog as it this is fixing something introduced in the last PR, which is described in the changelog.

### DIFF
--- a/corelib/src/libs/SireCAS/lambdaschedule.cpp
+++ b/corelib/src/libs/SireCAS/lambdaschedule.cpp
@@ -238,10 +238,12 @@ QString LambdaSchedule::toString() const
         auto keys = this->stage_equations[i].keys();
         std::sort(keys.begin(), keys.end());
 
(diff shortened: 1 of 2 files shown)


lohedges:
Thanks for catching this. Looks good :+1:

PR #162, opened 2024-02-11
Feature minimise

chryswoods:
Optimised the OpenMM minimisation code and making it more robust.
This includes vectorising for Apple Silicon and adding more tests for
convergence so that we can have more confidence that the structures
output are sensible. Also made sure that optimised compilation (-O3) is
used for all of the plugins (SireOpenMM, SireGemmi and SireRDKit).
They were previously compiled with wrapper options (e.g. -Os).
Minimisation now gives better progress updates, using a progress
bar to show progress towards the maximum number of iterations.
This has been reduced to 1500 by default. Also, if the minimisation
fails to create a structure that obeys constraints on the first pass,
then the minimisation is repeated, with the maximum number of
iterations reset. If it fails again, then this structure, with
constraints re-applied, is returned.

* I confirm that I have added a test for any new functionality in this pull request: [n/a (existing tests validate this)]
* I confirm that I have added documentation (e.g. a new tutorial page or detailed guide) for any new functionality in this pull request: [n/a (existing documentation describes this)]

## Suggested reviewers:
@lohedges, @mb2055 

It would be worth checking that minimisation is stable for a number of systems before accepting this PR. This makes the QM/MM alanine dipeptide system stable ("ala_25_0.rst7", "ala.top") when minimised. But, I have noticed that some TYK2 systems can be difficult to minimise (the algorithm has to try several times, and you can see progress pausing as the line search evaluates a lot without moving forwards)

This includes the LambdaSchedule bug fix from the other PR.

### DIFF
--- a/corelib/src/libs/SireBase/progressbar.cpp
+++ b/corelib/src/libs/SireBase/progressbar.cpp
@@ -1158,6 +1158,11 @@ static void check_raise_interrupt()
     }
 }
 
+void ProgressBar::silentTick()
+{
+    check_raise_interrupt();
+}
+
 void ProgressBar::tick()
 {
     check_raise_interrupt();
--- a/corelib/src/libs/SireBase/progressbar.h
+++ b/corelib/src/libs/SireBase/progressbar.h
@@ -86,6 +86,7 @@ namespace SireBase
 
         void tick();
         void tick(const QString &text);
+        void silentTick();
 
         void setProgress(quint32 value);
         void setProgress(quint32 value, const QString &text);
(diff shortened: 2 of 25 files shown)


lohedges:
Thanks, this sounds great. Looking at the actions there appear to be some build issues. I'll test from the QM/MM side once I've merged `devel` across to my branch. I just need to resolve conflicts and switch to the new auto-generation of wrappers. I'll let you know if I have any questions.

chryswoods:
That's brilliant - I've played some more and now have a good handle on the error codes that the underlying algorithm spits out. These tell you when you need to change algorithm parameters (e.g. increase line search steps etc). I am adding in some logic to handle these automatically, which is helping deal with the troublesome tyk2 cases. I'm going to add a bit more robustness to this PR before I merge it. My goal is to get the minimisation to work for everything except truly pathological cases ;-)

chryswoods:
I'll close this now so I don't accidentally trigger CI, and will reopen when I've finished.

Issue #170, opened 2024-02-23
[BUG] - Iron not recognised as a biological element


chryswoods:
Iron is not recognised as a biological element.

```
import sire as sr
fe = sr.mol.Element("Fe")
print(fe.biological())
```

prints `False`. This leads to iron atoms read from files that don't include element information to be incorrectly interpreted to be fluorine.

PR #171, opened 2024-02-23
fix_170 - detecting iron as biological

chryswoods:
This pull request fixes issue #170

Instead of special-casing iron, I've added functions to let the user control what elements are considered biological. I also took to opportunity to add easier conversions from strings, plus comparisons, so you can write ``Element("Fe") == "Fe"`` or ``atom.element() == "H"``.

## Suggested reviewers:
@lohedges

In writing the test I saw that running ``import sire as sr`` followed by ``sr.mol.Element`` raised an import error because of circular imports. Some of the changes here fix the circular import.


lohedges (approved):
All good. I've tested locally and everything is working as expected.

PR #187, opened 2024-04-16
Fix issue #186

### LINKED ISSUE #186: [BUG] sire.morph.create_from_pertfile doesn't assign different element properties to the end states
Not really a bug, since this works for the purposes of running simulations with `somd2`, which is the intended use case, but the `sire.morph.create_from_pertfile` function doesn't assign different elements to the end states when atoms are perturbed. (The `element0` and `element1` properties are identical.) This is because `element` is absent from the perturbation template and atom changes are inferred from the `ambertype0` and `ambertype1` properties.

This is currently causing round-trip pert file conversion to fail, since the file written back by BioSimSpace differs since it uses the element property to check for dummies, which will be wrong if there is a dummy in the lambda = 1 state. A minimal fix would be to set the appropriate end state property to `Element(0)` if either `ambertype0` or `ambertype1` is `du`. A full fix would be to correctly set the lamba = 1 element property for _all_ atoms, potentially inferring from `ambertype1`.

lohedges:
This PR fixes #186 by updating `sire.morph.create_from_pertfile` so that the `element1` property of the perturbable molecule is correct. Here, the property is inferred from the `ambertype1` property using `Element.biological_element`. I've added a test to make sure that the property is set correctly.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -16,6 +16,7 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 -----------------------------------------------------------------------------------------
 
 * Please add an item to this changelog when you create your PR
+* Correctly set the ``element1`` property in ``sire.morph.create_from_pertfile``.
 
 `2024.1.0 <https://github.com/openbiosim/sire/compare/2023.5.2...2024.1.0>`__ - April 2024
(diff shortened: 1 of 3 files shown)


chryswoods (approved):
👍

PR #188, opened 2024-04-21
Feature trajectory cache

chryswoods:
Added a `sire.base.PageCache` class which can be used to cache and restore objects to memory pages which are automatically paged to and from disk as needed. This lets you work on data that can't fit in memory.

Updated the way that `sire.system.System` objects hold the set of temporary frames in a trajectory. Rather than each molecule holding its own temporary frame, now the `System` object holds a ``SystemTrajectory`` object. This holds the frame data for all molecules in the `System` in a single binary array. The data for this array is paged to disk as needed via the above `sire.base.PageCache` class. This both significantly speeds up processing of these temporary frames, and ensures that long simulations do not fill memory, causing the system to crash. In addition, the ``SystemTrajectory`` object is NOT streamed to a S3 file. This means that the S3 file (used normally for restarts) won't grow unbounded with temporary frames, meaning that it is safe to create restarts of long-running simulations. Note that this does mean that the temporary directory is lost. You **must** save the trajectory to a file at the end of your simulation or it will be lost. You can do this using the standard trajectory save functions, e.g. ``sire.save(mols.trajectory(), "output", format=["PRMTOP", "RST"])``.

This should make trajectory management in somd2 much easier :-)

## Suggested reviewers:
@lohedges


lohedges:
Fantastic, thanks for this. I should be able to take a proper look on Weds. Related to this, does this also avoid the issue with the DCD file remaining open?} (This is currently causing test failures on WIndows, e.g. [here](https://github.com/OpenBioSim/somd2/actions/runs/8599509082/job/23562562549) but shouldn't cause an issue running simulations.)

chryswoods:
It should do - you can now avoid having to write the DCD files as chunks, and instead write at the end of the simulation if you want.

Or, if you still want chunks, we can look into using the new code in `SireBase/atexit.h`/`SireBase/atexit.cpp` to register exit handlers in C++ that will remove the files. It may be that this would be too late for the error you see though, as the exit handlers would be called when the Python script exits, while your error is raised when the Python temporary directory object is being destroyed.

chryswoods:
I've merged in your fix branch, and then killed GH Actions so that we don't waste cycles.

lohedges:
All is fine with `somd2`. For now I'll leave the code as-is since the chunking allows a user to recover the trajectory when starting from a checkpoint or crash. I'll wait to see if there are any reports of issues on Windows before exploring the new `atexit` handlers. It might not be a problem with the normal Python kernel teardown.

I'm currently seeing a single test failure in the BioSImSpace suite that I'm in the process of investigating. This test is [here](https://github.com/OpenBioSim/biosimspace/blob/9b605e8b33e2d900951e879d9cb105286313c8b5/tests/Process/test_somd.py#L42-L51) and the failure occurs because `None` is returned by `getSystem()` at the end of the simulation, presumably because it fails to read the trajectory file. This only happens for `somd-freenrg`, i.e. all of the other protocols (which use regular `somd`) appear to be unaffected. I'll run the test in a local working directory to figure out what's going wrong.

lohedges (approved):
No problems on my local testing, other than those caused by my own mistakes ;-)

PR #195, opened 2024-04-30
Fix issue #194

### LINKED ISSUE #194: [BUG] GroTop parser fails when position restraint include directives are present
The presence of include directives such as the following cause the parser to fail with no informative error.

```
#include "posre_0001.itp"
```

These files are written by BioSimSpace for when setting up a GROMACS system using position restraints. As such, the system can't be recovered from the topology and coordinate files without first removing these directives.

We could probably ignore this while parsing, but I'm not sure of the best approach. The `itp` extension is used for other includes too, so we can't just ignore that. BioSimSpace _does_ always write using the same format, i.e. `#include "posre_XXXX.itp``, so we could probably just match `posre` in the include?

I'll open a fix branch for this.

lohedges:
This PR closes #194 by skipping BioSimSpace position restraint include directives when parsing GROMACS topology files. I'm not sure if there is a more general way of fixing this issue, but this will work for the purposes of being able to read GROMACS files that were written as input to a `BioSimSpac.Process.Gromacs` simulation using position restraints. Here the restraints will be ignored, but they would be re-written if the user wanted to run the same protocol. In this instance, they just want to be able to read the system back in, i.e. the topology. 

## Suggested reviewers:
@chryswoods

### DIFF
--- a/corelib/src/libs/SireIO/grotop.cpp
+++ b/corelib/src/libs/SireIO/grotop.cpp
@@ -5399,6 +5399,12 @@ QVector<QString> GroTop::preprocess(const QVector<QString> &lines, QHash<QString
             }
         }
 
+        // skip BioSimSpace position restraint includes
+        if (line.contains("#include \"posre"))
+        {
+            continue;
+        }
+
         // now look for #include lines
         if (line.startsWith("#include"))
         {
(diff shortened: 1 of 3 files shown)


chryswoods (approved):
Are these restraints used only by BioSimSpace? If so, then I am happy to approve. Just want to check that we aren't ignoring something that another program would expect us to read.

lohedges:
This is the standard way of adding position restraints in GROMACS topology files, so similar directives might be present in other files too (which would also cause the parser to fail). Here I've explicitly skipped the records when the exact BioSimSpace naming convention is found, but I could also update that to be more unique, e.g. include bss in the string.

Another solution would be adding support for parsing these records, but I'm not sure how useful that would be.

chryswoods:
Ok - we can use this skip for now, while I think about how to add support for parsing the record. It may be useful to have these as extra properties of the molecule or system.

Issue #208, opened 2024-07-10
Improve logging of co-alchemical water perturbations in SOMD1


jmichel80:
SOMD1 currently silently picks water molecule(s) for perturbation into a sodium or chloride ion when a charge changing ligand perturbation is attempted with PME.  

We should log the identify of the water molecules to facilitate troubleshooting and optimisation of the simulation protocol.

lohedges:
Closed via #209.

PR #209, opened 2024-07-10
now printing to stdout the identify of the water molecules that are p…

jmichel80:
now printing to stdout the identify of the water molecules that are perturbed into ions if a co-alchemical water perturbation has been activated with SOMD1

## If this is to fix a bug...

This pull request fixes issue #208 

## Suggested reviewers:
@chryswoods, @lohedges

### DIFF
--- a/wrapper/Tools/OpenMMMD.py
+++ b/wrapper/Tools/OpenMMMD.py
@@ -2505,6 +2505,7 @@ def selectWatersForPerturbation(system, charge_diff):
         # FIXME: select waters according to distance criterion
         # if mol.residue().name() == water_resname and cnt < nions:
         if mol.residues()[0].name() == water_resname and cnt < nions:
+            print ("Selected water %s for perturbation into ion" % mol)
             cnt += 1
 
             perturbed_water = mol.edit()
--- a/wrapper/Tools/OpenMMMD.py
+++ b/wrapper/Tools/OpenMMMD.py
@@ -2505,7 +2505,7 @@ def selectWatersForPerturbation(system, charge_diff):
         # FIXME: select waters according to distance criterion
         # if mol.residue().name() == water_resname and cnt < nions:
         if mol.residues()[0].name() == water_resname and cnt < nions:
-            print ("Selected water %s for perturbation into ion" % mol)
+            print ("Selected water residue %s for perturbation into ion" % (mol.residues()[0]))
             cnt += 1
 
             perturbed_water = mol.edit()


wrapper/Tools/OpenMMMD.py
```diff
@@ -2505,6 +2505,7 @@ def selectWatersForPerturbation(system, charge_diff):
         # FIXME: select waters according to distance criterion
         # if mol.residue().name() == water_resname and cnt < nions:
         if mol.residues()[0].name() == water_resname and cnt < nions:
+            print ("Selected water %s for perturbation into ion" % mol)
```
lohedges, line 2508: This would just print the molecule, so would display something like:

```python
Molecule( WAT:477 num_atoms=3 num_residues=1 )
```

Given that waters have the same naming, and the ordering `MolNum` (here 477) doesn't correspond to the index wsithin the molecule, I wonder if it would be more useful to print the water index from the PRM7 topology that was loaded?

For example, if I load test files in sire I find:

```python
In [1]: import sire as sr

In [2]: mols = sr.load_test_files("ala.top", "ala.crd")

In [3]: mols.num_molecules()
Out[3]: 631

In [4]: mols[-1].number()
Out[4]: MolNum(421)
```

Here I have printed the `MolNum` of the last molecule, which doesn't correspond to the number of molecules in the system. If I print the number for the first 20 I get:

```python
In [5]: for mol in mols[:20]:
   ...:     print(mol.number())
   ...:
MolNum(21)
MolNum(251)
MolNum(260)
MolNum(277)
MolNum(293)
MolNum(311)
MolNum(327)
MolNum(343)
MolNum(359)
MolNum(376)
MolNum(393)
MolNum(409)
MolNum(424)
MolNum(440)
MolNum(459)
MolNum(475)
MolNum(492)
MolNum(509)
MolNum(513)
MolNum(276)
```

As such, I think it would be hard for the user to cross reference the `MolNum` with the actual water that was perturbed in the topology. However, it does appear that printing by index might be a bit fiddly due to the way that you separate the system into different molecule groups, e.g. the solvent is in a separate group. At this point, the waters already aren't in the same order as the original topology due to the way the containers store items. (This is an issue that BioSimSpace deals with by mapping things back to the original topology and re-numbering residues so that they are guaranteed to be unique.) Things might be easier if perturbed waters had a unique residue name, but that might cause other issues.

jmichel80: A possible workaround is to print mols[X].residues()[0] which assumes that the residue number will match the ordering of the residues in the prm7 file.

lohedges: I think that should work. The `ResNum` will be preserved across container operations and will match what was loaded from the original system. This is what BioSimSpace uses to track and map things. (It first renumbers the entire system so that the residue numbers are unique and in ascending order.)

```python
In [1]: import sire as sr

In [2]: mols = sr.load_test_files("ala.top", "ala.crd")

In [3]: for mol in mols[:20]:
   ...:     print(mol[0].residues()[0].number())
   ...:
ResNum(1)
ResNum(4)
ResNum(5)
ResNum(6)
ResNum(7)
ResNum(8)
ResNum(9)
ResNum(10)
ResNum(11)
ResNum(12)
ResNum(13)
ResNum(14)
ResNum(15)
ResNum(16)
ResNum(17)
ResNum(18)
ResNum(19)
ResNum(20)
ResNum(21)
ResNum(22)
```

lohedges (approved):
Great, this works :+1:

PR #211, opened 2024-07-11
Fix AtomMapping::alignTo when mapping an ion to a molecule

lohedges:
This PR fixes the `AtomMapping::alignTo` methods to handle the case where one molecule is a monatomic ion. This allows `sire.morph.merge` to handle the creation of alchemical ions by merging a water molecule with an ion. The existing `alignTo` methods fail since they attempt a RMSD alignment when there are too few degrees of freedom.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/corelib/src/libs/SireMol/atommapping.cpp
+++ b/corelib/src/libs/SireMol/atommapping.cpp
@@ -31,6 +31,7 @@
 #include "SireMaths/align.h"
 
 #include "SireMol/core.h"
+#include "SireMol/moleditor.h"
 
 #include "SireStream/datastream.h"
 #include "SireStream/shareddatastream.h"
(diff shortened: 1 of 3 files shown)


lohedges:
Note that this is just assuming a single ion is mapped, which is the use case in question. I guess we could generalise this to fix when the `AtomMapping` contains multiple molecules that are mapped to ions.

lohedges:
I just need to add a platform skipif decorator for Windows. I'll add this tomorrow.

chryswoods (approved):
Thanks - nice fix :-)

lohedges:
Thanks. I'll merge this in so that we can work on the `somd2` alchemical ion implementation next week.

PR #214, opened 2024-07-24
Fix issue #213

### LINKED ISSUE #213: [BUG] OpenMMPMEFep wrappers aren't excluded when OpenMM is missing
We need to add the `OpenMMPMEFep` wrappers to `wrapper/Move/CMakeOpenMMFile.txt` to make sure that they are excluded when OpenMM isn't available. Currently it only has:

```
# WARNING - AUTOGENERATED FILE - CONTENTS WILL BE OVERWRITTEN!
set ( PYPP_OPENMM_SOURCES
       OpenMMMDIntegrator.pypp.cpp
       OpenMMFrEnergyDT.pypp.cpp
       OpenMMFrEnergyST.pypp.cpp
    )
```

(I'm not sure if this file was autogenerated since the only mention of the file name is within `wrapper/Move/CMakeLists.txt` itself.)

lohedges:
This PR closes #213 by excluding the OpenMMPMEFEP wrapper files when OpenMM is missing.

## Suggested reviewers:
@chryswoods


lohedges:
Actually, I also need to add a file to `wrapper/Move/NoOpenMM`. I didn't realise that there were wrapper stubs for the classes, even when OpenMM is missing. (In which case, the classes are useless, surely?)

lohedges:
The CI failures appear to be because RDKit wasn't installed for some reason, or a version that works with SireRDKit at runtime.

chryswoods (approved):
👍

Issue #217, opened 2024-07-25
Migrate rdkit-dev to librdkit-dev


skearnes:
Hi, this is a friendly notification that the `rdkit-dev` conda-forge package has been replaced by `librdkit-dev` as of https://github.com/conda-forge/rdkit-feedstock/pull/158; please update your dependencies.

@lohedges thanks for inspiration to do this via github code search :)

lohedges:
Many thanks. Currently working on this in #216.

skearnes:
Thanks; I'll reopen the issue until we get that resolved.

skearnes:
@lohedges could you send me a link to the CI log? If there's a specific header I can look for that would be helpful.

lohedges:
The last working Windows Python 3.12 log is [here](https://github.com/user-attachments/files/16409896/working.txt.gz). At build time it used:

```
2024-06-29T17:08:52.9966320Z     rdkit:                     2024.03.3-py312h7e6e52d_0    conda-forge
2024-06-29T17:08:52.9966882Z     rdkit-dev:                 2024.03.3-h36c15f3_0         conda-forge
```

The library and headers were correctly detected:

```
2024-06-29T17:38:32.1135002Z -- Looking for RDKit in C:/Users/runneradmin/miniconda3/envs/sire_build/conda-bld/sire_1719680378610/_h_env/incl>
2024-06-29T17:38:32.1139376Z -- Found RDKit libraries at C:/Users/runneradmin/miniconda3/envs/sire_build/conda-bld/sire_1719680378610/_h_env/>
2024-06-29T17:38:32.1150448Z -- Found RDKit library files at C:/Users/runneradmin/miniconda3/envs/sire_build/conda-bld/sire_1719680378610/_h_>
2024-06-29T17:38:32.1167780Z -- Found RDKit: C:/Users/runneradmin/miniconda3/envs/sire_build/conda-bld/sire_1719680378610/_build_env/Library/>
2024-06-29T17:38:32.1169742Z -- Compiling SireRDKit converter
```

For the tests it pulled in...

```
2024-06-29T18:51:06.3814956Z     rdkit:                     2024.03.3-py312h7e6e52d_0    conda-forge
```

and all the RDKit related tests passed, e.g.:

```python
2024-06-29T18:53:43.7385564Z tests/convert/test_rdkit.py::test_rdkit[C1CCCCC1] PASSED                 [ 15%]
[...]

PR #216, opened 2024-07-24
Fix issue #215

lohedges:
This PR closes #215 by setting the `IFBOX` pointer to 3 for triclinic spaces. The value of 2 is reserved for a truncated octahedron. I'm not sure if we should add logic to detect this, or whether the use of 3 should suffice for everything. I'll test and report back if there are any issues.

Note that I've killed this superses #214, so I've killed the CI for that PR.

This also closes #217 by switching to the `librdkit-dev` package as a replacement for the now deprecated `rdkit-dev`.

* I confirm that I have merged the latest version of `devel` into this branch before issuing this pull request (e.g. by running `git pull origin devel`): [y]
* I confirm that I have added a changelog entry to the changelog (we will add a link to this PR as part of the review): [y/n]
* I confirm that I have permission to release this code under the GPL3 license: [y]

## Suggested reviewers:
@chryswoods

PR #220, opened 2024-07-29
Fix issues 213, 215, and 219

### LINKED ISSUE #219: [BUG] Undesired behaviour in systems with multiple perturbable molecules
**Describe the bug**
Recent PRs to sire, biosimspace and SOMD2 introduced the ability to include co-alchemical waters within systems with charge perturbations. In testing this functionality some undesirable behaviour has been encountered, specifically in the interactions between the co-alchemical waters and the atom within the ligand that is responsible for the charge change. This issue manifests as the co-alchemical water and charged atom becoming strongly bound over the course of the simulation, but only in intermediate states (behaviour at lambda=0 and lambda=1 is as expected). 

Plots of distance look as follows
prefix_dists

**To Reproduce**
Run the following simple simulation script:
```
import sire as sr

system = sr.load("STREAMED_SYSTEM.s3")
system = sr.morph.link_to_reference(system)
system = system.minimisation(lambda_value=0.5).run().commit()

d = system.dynamics(lambda_value=0.5, timestep="4fs", temperature="25oC")
d.run("5ns", save_frequency="1ps")
system = d.commit()
sr.save(system.trajectory(), "lambda05.dcd", format=["DCD"])
```

and look at the distances between the atom `DU02` within the ligand and the molecule at index `489`, this is the alchemical water.

**Expected behavior**
The distribution of distances between the co-alchemical ion and `DU02` should look largely the same, regardless of lambda value, and should not have the clear drops to zero seen in the trajectory above.

**Input files**
[input.zip](https://github.com/user-attachments/files/16415009/input.zip)

 - OS:  Ubuntu 22.04.4 LTS
 - Version of Python: 3.12.4
 - Version of sire: 2024.3.0.dev
 - I confirm that I have checked this bug still exists in the latest released version of sire: yes

**Additional context**
The bug seems to originate from [this](https://github.com/OpenBioSim/sire/commit/44ce0e7cee9d2ed3d8a296fbfe13ae76f5f8df6d) PR, in which exclusions were added for from/to ghosts. Currently these exceptions are added for all to/from ghost atoms in the system, when they should only be added for atoms within the same molecule. 

@lohedges has a fix that ensures that these exceptions are added correctly. Testing of the fix confirms that it fixes the issue:
postfix_dists

lohedges:
This PR closes #219 by only excluding non-bonded interactions between from/to ghost atoms within the _same_ molecule. This is one of several possible solutions (probably) the easiest. The issue in #219 can also be avoided by using a distance restraint between the alchemical ion and perturbable molecule, but this isn't used in `somd1` and the implementation works without issues. Regardless, I think that the logic to exclude the non-bonded interactions was only intended for atoms within a perturbable molecule anyway.

In the fix I've just reused the existing `start_indices` vector to work out the molecule index. I'm happy to also create a `atom_to_mol` container or similar to avoid the double molecule index search. (This was just a quick fix, re-using what we had.)

Note that I've currently skipped the CI. I can trigger once we resolved issue #217. This PR supersedes #216.

## Suggested reviewers:
@chryswoods


chryswoods:
This looks good - I am happy with the code you've used to detect molecules. It is very simple. We can look to optimise it in the future if needed. I guess this would only possibly impact things when we have, e.g. perturbable proteins?

chryswoods (approved):

lohedges:
Yes, I doubt it's much of a performance issue. If needed, it's trivial to keep track of the molecule indices as the ghosts are detected. I just did this as the simple first fix and thought I'd optimise once we knew it fixed the problem. (And if optimisation was needed.)

PR #221, opened 2024-07-29
Use old rdkit-dev package in SireRDKit build

### LINKED ISSUE #217: Migrate rdkit-dev to librdkit-dev

skearnes:
Related to #217

### DIFF
--- a/requirements_build.txt
+++ b/requirements_build.txt
@@ -13,8 +13,7 @@ sysroot_linux-64==2.17 ; sys_platform == "linux"
 
 # These packages are needed to compile
 # the SireRDKit plugin
-rdkit >=2023.0.0
-rdkit-dev >=2023.0.0
+rdkit-dev >=2024.03.05
 
 # These packages are needed to compile
 # the SireGemmi plugin
--- a/requirements_build.txt
+++ b/requirements_build.txt
@@ -14,7 +14,7 @@ sysroot_linux-64==2.17 ; sys_platform == "linux"
 # These packages are needed to compile
 # the SireRDKit plugin
 # TODO(skearnes): Remove this pin; see https://github.com/conda-forge/rdkit-feedstock/issues/167.
-rdkit-dev ==2024.03.04
+rdkit-dev <=2024.03.04
 
 # These packages are needed to compile
 # the SireGemmi plugin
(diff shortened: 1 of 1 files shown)


lohedges:
Nope, still no joy. I'll just use `librdkit-dev` as I have in #220. We can try to figure out the Windows wrapper issue in time for the next release.

skearnes:
There's something deeper going on here if that pin didn't work. That should
have taken us back to before any of the recent changes.

lohedges:
I think I see the issue. For some reason (perhaps @chryswoods can comment) we originally needed to included `rdkit` and `rdkit-dev` as in the `build` section in order to get the headers and libraries detected on all platforms. Our `host` section also includes dependencies that are required by tools built on top of Sire, so that things can be built in a self-consistent way. This _also_ includes `rdkit`, via BioSimSpace's depdencies. This requirement doesn't share the same pin as the one in `build`, so it seems to be pulling in the most recent version, i.e. without the headers. For example, for Linux Python 3.12:

```
rdkit:                            2024.03.5-py312h7b4b7d0_2              conda-forge 
```

I'll adjust my most recent PR to just use just `rdkit<=2024.03.4` and `rdkit-dev<=2024.03.4` in the `host` section only.

Weirdly, having just checked, installing `rdkit=2024.03.4` alone is enough to bring in the headers, so perhaps `rdkit-dev` isn't even needed.

lohedges:
Yes, that fixed things. The issue here was that the `rdkit` requirement was being defined in _two_ places, with different pins. The latter took precedence, which broke the pin that you introduced. I'll close this since I now have it working in #220. I'll try once more just using the new `librdkit-dev` instead to see if removing the duplicate fixes the issues there.

chryswoods:
Sorry - I think that was historic to get things working in the past on different OS's. It can be removed if it isn't needed any more.

lohedges:
That has also fixed the weird boost wrapper issue with the Windows build when using `librdkit-dev`. Phew. Think we got there in the end. Apologies for all the confusion.

skearnes:
No worries! Glad we got it sorted.

PR #222, opened 2024-07-29
Fix missing is_mechanical from (de)serialization and assignment opera…

JMorado:
This PR addresses a bug where information about the embedding scheme was lost, causing Sire to assume that mechanical embedding was being used. Consequently, the neighbors list was not calculated when it should have been. This issue occurred when using the [`get_forces` method](https://github.com/JMorado/sire/blob/feature_emle/src/sire/qm/_emle.py#L41), as deepcopying serialized and deserialized the `PyQMForce` object, leading to the loss of the embedding scheme information.

* I confirm that I have merged the latest version of `devel` into this branch before issuing this pull request (e.g. by running `git pull origin devel`): 

## Suggested reviewers:
@lohedges


lohedges:
Thanks for catching this.

PR #223, opened 2024-08-05
Feature linux x86 wrappers

lohedges:
This PR adds a Dockerfile for building the Sire wrappers on Linux x86, called `Dockerfile_x86`.  (I've renamed the existing Dockerfile as `Dockerfile_arm64` and updated the build instructions accordingly.)

I've also added a `--lib` option to the `generate_wrappers` script to allow a user to only build the wrappers for a specific library from corelib, rather than all of the wrappers. This significantly reduces the build time when development is contained to one or few libraries.

Note that some of the dependencies in the container are quite old, e.g. Python 3.7 and clang10. This is all that I could get working locally, but will explore ways to update in future, e.g. retrying, or patching as necessary.

## Suggested reviewers:
@chryswoods


chryswoods (approved):
Thanks - great that we have a working Linux X86-64 container again :-)

PR #229, opened 2024-08-19
Fix #228

### LINKED ISSUE #228: [BUG] Issues with dummy atomtypes in perturbable GROMACS topologies
I'm not sure how this hasn't been triggered before, but there appear to be some issues with the `[ atomtypes ]` and `[ atoms ]` directives for perturbable GROMACS topologies. In particular, the `[ atoms ]` section can contain duplicate `_du` suffixes, so the types don't always match those in the `[ atomtypes ]` section. There also appear to be some types that aren't used in the `[ atoms ]` section, suggesting that the deduplication code isn't working as expected.

lohedges:
This PR closes #228 by correcting the logic to set GROMACS atom types for dummy atoms. In certain cases an additional `_du` suffix was applied to dummies when writing to the `[ atoms ]` section, meaning that no match was found in the `[ atomtypes ]` section. In addition, atom type deduplication wasn't always working correctly meaning that there were (sometimes) unused atom types in the `[ atomtypes ]` section, i.e. they never appeared in `[ atoms ]`. This meant that an atom in `[ atoms ]` may have been paired with the wrong type from `[ atomtypes ]`. (This doesn't matter for all terms, since those in `[ atoms ]` take precedence.)

I'm still not sure how this issue hasn't occurred before. It's not clear whether this was new input for Exscientia, or whether an existing test triggered the problem following the LJ sigma update. (Although I can't see how.)

I've added a unit test to validate the two sets of records for the problem system and have also checked that all of the BioSimSpace unit tests still pass.

Note that I'd probably like to go back and simplify the logic here at some point. This is a quick fix to get things working before I go on holiday.

## Suggested reviewers:
@chryswoods

### DIFF
--- a/corelib/src/libs/SireIO/grotop.cpp
+++ b/corelib/src/libs/SireIO/grotop.cpp
@@ -3217,14 +3217,6 @@ static QStringList writeMolType(const QString &name, const GroMolType &moltype, 
                     elem1 = Element::elementWithMass(mol.property("mass1").asA<AtomMasses>()[cgatomidx]);
                 }
 
-                // Update the atom types.
-
-                if (elem0.nProtons() == 0)
-                    atomtype0 += "_du";
-
-                if (elem1.nProtons() == 0)
-                    atomtype1 += "_du";
-
                 QString resnum = QString::number(atom0.residueNumber().value());
 
                 if (not atom0.chainName().isNull())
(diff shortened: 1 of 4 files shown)


xiki-tempula:
Do you mind make a branch targeting main so I could have a test? Thanks.

lohedges:
[Here](https://github.com/OpenBioSim/sire/tree/backport_229) we go.

lohedges:
@xiki-tempula: Did you get a chance to test this? If all is okay, then I'll merge into `devel` this week and backport.

xiki-tempula:
Yes. It works very well.

lohedges:
Perfect. Thanks for confirming.

PR #233, opened 2024-09-26
Fix #232

### LINKED ISSUE #232: [BUG] Minimisation/Dynamics failing when using positional restraints
**Describe the bug**
Simulations fail when running  minimisation or dynamics on a perturbable system with positional restraints. 

**To Reproduce**
Run the following script:
```python
import sire as sr
system = sr.load(sr.expand(sr.tutorial_url, "merged_molecule.s3"))
system = sr.morph.link_to_reference(system)
restraints = sr.restraints.positional(system, atoms="element C")
system = system.minimisation(lambda_value=0.0, restraints=restraints).run().commit()
```
The error raised is: `ValueError: There was a problem creating the OpenMM context. Perhaps the platform was not supported for this system, options or on this computer? The error message is: CustomNonbondedForce: Wrong number of parameters 
for particle 12167`

**Expected behavior**
The script should minimise as expected, which it does if the `restraints` argument is removed.

 - OS: Ubuntu 22.04.5 LTS
 - Version of Python: Python 3.12.4
 - Version of sire: 2024.3.0.dev
 - I confirm that I have checked this bug still exists in the latest released version of sire: yes

**Additional context**
The system in question contains exactly 12167 atoms. This implies that the issue here is in the additional ghost atoms added when using positional restraints.

mb2055:
This PR closes #232. Changes the length of the parameter vector used to add ghost atoms (specifically those used for positional restraints) from 3 to the required length of 5.
* I confirm that I have permission to release this code under the GPL3 license: [y/n]

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -26,7 +26,7 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 * Add support for boresch restraints to PME.
 * Port SOMD torsion fix to PME code.
 * Fix issues with ``atomtype`` and ``atom`` records for dummy atoms in GROMACS topology files.
-* Fix issues with positionally restrainted atoms in perturbable systems.
+* Fix issues with positionally restrained atoms in perturbable systems.
 
 
 `2024.2.0 <https://github.com/openbiosim/sire/compare/2024.1.0...2024.2.0>`__ - June 2024
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -26,6 +26,7 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 * Add support for boresch restraints to PME.
 * Port SOMD torsion fix to PME code.
 * Fix issues with ``atomtype`` and ``atom`` records for dummy atoms in GROMACS topology files.
+* Fix issues with positionally restrainted atoms in perturbable systems.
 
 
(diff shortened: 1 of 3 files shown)


doc/source/changelog.rst
```diff
@@ -26,6 +26,7 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 * Add support for boresch restraints to PME.
 * Port SOMD torsion fix to PME code.
 * Fix issues with ``atomtype`` and ``atom`` records for dummy atoms in GROMACS topology files.
+* Fix issues with positionally restrainted atoms in perturbable systems.
```
lohedges, line 29: Typo: `resttrainted` should be `restrained`.

lohedges (changes requested):
Thanks for this. Is it possible to add a unit test for this. Probably within `tests/convert/test_openmm_restraints`. Youl could follow the logic for the existing `test_openmm_positional_restraints` function, but instead using one of the perturbable molecule fixtures, e.g. `merged_ethane_methanol`. You could also parameterise the test so the `mols` is `kigaki_mols` _and _ `merged_ethane_methanol`. There are tests in the BioSimSpace suite that show how to do this.

lohedges (approved):
Thanks for the edits. I'll combine this with the other fix PRs then merge in one go once we're happy with everything.

PR #237, opened 2024-09-27
Fix #236

### LINKED ISSUE #230: [BUG] Minimiser hangs
In quite a few instances I am seeing hangs with the modified OpenMM minimiser. I've tried to mitigate this using the `max_iterations` kwargs, but the problem persists. It seems that the number of iterations might be set to zero for each ratchet. The available options are:

```
    Parameters:

    - max_iterations (int): The maximum number of iterations to run
    - tolerance (float): The tolerance to use for the minimisation
    - max_restarts (int): The maximum number of restarts before giving up
    - max_ratchets (int): The maximum number of ratchets before giving up
    - ratchet_frequency (int): The maximum number of steps between ratchets
    - starting_k (float): The starting value of k for the minimisation
    - ratchet_scale (float): The amount to scale k at each ratchet
    - max_constraint_error (float): The maximum error in the constraint in nm
```

If it's not possible to set things independently, i.e. an absolute number of iterations as the max, then maybe we also want to implement a timeout of some sort? I'll try to find a system that repeatedly hangs to post for debugging purposes. That's one of the other things that's a bit frustrating, since this doesn't happen _all_ the time, e.g. you might see it for one window of one replica of an RBFE run, which causes the whole thing to fail.

### LINKED ISSUE #236: [BUG] Buffer overflow causes inconsistent number of ghost-ghost exclusions
When working out the molecule indices for the from-ghost and to-ghost atoms to exclude interactions between them (when they are in the same molecule) the loop indexes past the end of the `start_indexes` vector, causing memory corruption. This leads to an inconsistent number of exclusions in the force field from run-to-run.

### LINKED ISSUE #238: [BUG] Incorrect calculation of soft-core Coulomb delta parameter
According to the documentation [here](https://sire.openbiosim.org/versions/devel/tutorial/part07/03_ghosts.html#zacharias-softening), the `delta` parameter used in the `Nonbonded` and `CustomNonbonded` forces is defined as `delta = alpha * shift_coulomb`, so `delta^2 = alpha^2 * shift_coulomb^2`. The value of `shift_coulomb` is passed to the force already squared, then is simply multiplied by `alpha` (or `max_alpha`) in the expression to generate `delta^2`. This should instead be multiplied by `alpha^2`.

### LINKED ISSUE #239: [BUG] to/from ghost interactions aren't excluded from the ghost_14 force field
We have modified the code to add exclusions for to/from ghost atom interactions from the ghost/ghost `CustomNonbondedForce`. These interactions also need to be excluded from the ghost_14 `CustomBondForce`.

### LINKED ISSUE #241: [BUG] Incorrect delta energy evaluated following constraint projection after energy minimisation
The OpenMM energy minimiser doesn't re-calculate the system energy following the constraint projection performed after energy minimisation. This means that the delta energy used to compute the change in energy due to application of constraints is incorrect.

### LINKED ISSUE #242: [BUG] self._omm_state isn't set to None after minimisation
The internal OpenMM state of the dynamics object needs to be cleared following a minimisation so that calling `d.current_potential_energy()` or similar will get the updated state before computing the energy.

lohedges:
This PR closes #236 by restricting the loop over the `start_indexes` vector to the number of molecules in the system.

Also closes #238 by fixing the calculation of `delta^2` used in the Coulomb term of the nonbonded expressions.

Also closes #239 by excluding to/from ghost interactions from the `ghost_14ff`.

Also closes #230 by adding a crude timeout to the OpenMM minimiser. This just checks to see if a maximum number of seconds has been exceeded in both the inner and outer loops of the code. This obviously doesn't check the timing within the internal line search algorithm itself. The main point of this is allow the user a single lever to control the max time (roughly) before the minimiser exits, rather than needing to adjust the number of iterations, ratchets, etc. (And even then, those might still lead to a hang.)

Also closes #241 by re-computing the energy _after_ performing the final constraint projection in order to get the correct delta energy.

Also closes #242 by clearing the internal OpenMM state of the dynamics object when performing a minimisation.

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -29,6 +29,8 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 * Fixed buffer overflow when computing molecule indices to excluded to/from ghost atom interactions.
 * Fixed calculation of ``delta^2`` in soft-core Couloumb potential.
 * Excluded to/from ghost atom interactions from ``ghost_14ff``.
+* Fixed description of soft-core alpha parameter in tutorial.
+* Added debugging function to evaluate custom forces in OpenMM XML files.
 
 
 `2024.2.0 <https://github.com/openbiosim/sire/compare/2024.1.0...2024.2.0>`__ - June 2024
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -34,7 +34,7 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 * Added a timeout to the OpenMM minimiser function.
 * Exposed the pickle operator on the LambdaLever class.
 * Fix issues with positionally restrained atoms in perturbable systems.
-
+* Fix exchange probability equations in ``sire.morph.replica_exchange`` function.
 
 `2024.2.0 <https://github.com/openbiosim/sire/compare/2024.1.0...2024.2.0>`__ - June 2024
 -----------------------------------------------------------------------------------------
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -35,6 +35,8 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 * Exposed the pickle operator on the LambdaLever class.
 * Fix issues with positionally restrained atoms in perturbable systems.
 * Fix exchange probability equations in ``sire.morph.replica_exchange`` function.
+* Fix calculation of energy change following final constraint projection after energy minimisation.
+* Clear internal OpenMM state from dynamics object following a successful minimisation.
 
 `2024.2.0 <https://github.com/openbiosim/sire/compare/2024.1.0...2024.2.0>`__ - June 2024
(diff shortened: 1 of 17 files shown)


mb2055 (approved):
:+1: All good

lohedges:
I've also added a new `sire.morph.evaluate_xml_force` function that can be used to get the atoms and energies involved in the custom forces from an XML file that we generate, e.g. something liike:

```python
In [1]: import sire as sr

In [2]: mols = sr.load_test_files("merged_molecule.s3")

In [3]: mols = sr.morph.link_to_reference(mols)

In [4]: d = mols.dynamics()

In [5]: from openmm import XmlSerializer

In [6]: with open("debug.xml", "w") as f:
   ...:     f.write(XmlSerializer.serialize(d._d._omm_mols.getSystem()))
   ...:

In [7]: pairs, nrg_coul, nrg_lj = sr.morph.evaluate_xml_force(mols, "debug.xml", "ghost-14")
```

PR #243, opened 2024-10-10
Add support for QM/MM

### LINKED ISSUE #244: [BUG] Constraints in OpenMM context aren't updated
Calling `context.getSystem().setConstraintParameters` doesn't update the parameters in the context that holds the system, hence calling `d.set_lambda(lam_val, update_constraints=True)` doesn't actually update the constraints. This is the cause of the hysteresis in ddG values for alchemical perturbations with different constraints at the end states.

lohedges:
This PR adds support for QM/MM simulations in Sire using OpenMM as the MM backend. Though the interface is primarily designed to facilitate support for ML/MM simulations using [emle-engine](https://github.com/chemle/emle-engine) it is completely general and can be used to interface with _any_ external QM engine, or even to implement any custom time-dependent forced based on the local environment around some region.

Available QM engines are:

* `PyQMEngine`: This is a general purpose Python callback that can be used to quickly prototype QM/MM code, or to interface with more complex external tools, e.g. those without an C++ or Python API, such as command-line executables.
* `TorchQMEngine`: This provides direct support for Torch based ML models. Since Torch isn't (yet) available on WIndows via conda-forge, support for this engine is conditionally compiled.

The above are exposed to the user using the new Sire Python API via the `sire.qm` sub-module. Convenience functions, such as `sire.qm.create_engine` and `sire.qm.emle` simplify the creation of QM engines for running QM/MM dynamics with an existing Sire dynamics object.

QM molecules are treated with the same "perturbable" molecule framework that is used for alchemical free-energy simulations. This allows for interpolation between MM and QM potentials via a "lambda" parameter, which can be used for, e.g., end-state correction simulations.

The code also provides support for using EMLE with the existing OpenMM-ML framework, i.e. you can simply obtain the force used to compute the electrostatic embedding interaction, then add it to an existing OpenMM-ML mixed potential system.

As for BioSimSpace, an additional set of conditional requirements are defined in the `requirements_emle.txt` file, which allows Sire to be built in a environment that is self-consistent with the requirements of `emle-engine`. Rather than automatically including these requirements for all builds, I decided to expose them via the `SIRE_EMLE` environment variable. This allows us to easily disable the requirements, e.g. if we find conflicts for specific platforms, or if the additional requirements lead to unwanted constraints on the environment for other users, e.g. our industrial collaborators.

Note that I haven't selectively checked for `CustomCPPImpl` support. The logic doesn't work locally for me on Linux, so I needed to remove from my local repository. For the CI, we only ever build for version 8.1+, so this isn't an issue. We could re-instate the logic if you worry that someone might try to do a source build against an older OpenMM, although we should try to figure out why the logic fails. (I see the same problem for `updateSomeParametersInContext` too, but that will soon be redundant anyway.)

This PR also closes #244 by reinitialising the OpenMM context if constraints change when setting lambda. This is required for the constraint change to propagate context itself, i.e. setting the parameters in the system within the context is not enough. This is quite painful, since it's an expensive operation. Thankfully constraints are only set to lambda specific values at the outset of a simulation and aren't updated when computing energies at other lambda values. This could cause issues for multistate sampling methods, such as HREX, though, where we _would_ need to update the constraints when swapping lambda. If this becomes a problem then we can look deeper into OpenMM itself to see if there are optimisations to make.

## Suggested reviewers:
@chryswoods


lohedges:
One thing to note is that the instructions for creating the `sire-emle` conda environment in the tutorial assume the default `main` branch of `emle-engine`, which sill uses the old SOAP based models. Once this PR is done, I'll bring the `feature_aev` branch of `emle-engine` into `main` so that everything will be in sync.

lohedges:
I've modified the callback signature to accept an optional list of MM indices as the last parameter. This will allow a user to get any additional atomic properties that they might need in their callback, while also keeping things backwards compatible.

Annoyingly it appears that the OpenMM XMLSerializer (or whatever it uses as the backend) doesn't preserve precision on macOS, so you can't sort and diff files as a test. I'll restrict this to the platforms where it works, i.e. Linux, since it's a useful regression test to have.

doc/source/tutorial/part08/02_emle.rst
```diff
@@ -0,0 +1,435 @@
...
+We can now run the simulation. The options below specify the run time, the
+frequency at which trajectory frames are saved, and the frequency at which
+energies are recorded. The ``energy_frequency`` also specifies the frequency
+at which the λ value is updated.
+
+>>> d.run("1ps", frame_frequency="0.05ps", energy_frequency="0.05ps")
+
+.. note::
+
+    Updating λ requires the updating of force field parameters in the ``OpenMM``
```
chryswoods, line 98: Do we still need to mention this now that the new version of OpenMM is out with the optimised updateParametersInContext?

doc/source/tutorial/part08/04_diels_alder.rst
```diff
@@ -0,0 +1,251 @@
...
+)
+
+.. note::
+
+    In ``OpenMM``, fixed atoms are implemented by setting atomic masses to zero.
+    This means that the atoms are still involved in interactions, but do not move.
+
+Now we will extract the context from the dynamics object:
+
+>>> context = d._d._omm_mols
```
chryswoods, line 122: Would it be worth adding a `d.extract_context()` function (or similar) that would make getting the context a more "official" operation? The tutorial shows that this is a very useful thing to do. It is a bit messy to do this by calling private functions.

wrapper/Convert/SireOpenMM/lambdalever.cpp
```diff
@@ -1891,13 +1912,13 @@ void LambdaLever::updateRestraintInContext(OpenMM::Force &ff, double rho,
     // what is the type of this force...?
     const auto ff_type = ff.getName();
 
-    if (ff_type == "CustomBondForce")
+    if (ff_type == "BondRestraintForce" or ff_type == "PositionalRestraintForce")
```
chryswoods, line 1915: Nice renaming of the custom forces - I like all the new names

doc/source/tutorial/part08/02_emle.rst
lohedges: Good point, this was more for Exs' benefit so can be safely removed.

doc/source/tutorial/part08/04_diels_alder.rst
```diff
@@ -0,0 +1,251 @@
...
+>>> mols = sr.load_test_files("abyu.prm7", "abyu.rst7")
+
+Next we use a ``sire`` `selection <https://sire.openbiosim.org/versions/devel/cheatsheet/search.html>`_
+to create a sphere around the reaction site:
+
+>>> water_sphere = mols["water within 22A of atomidx 3 in molidx 1"]
+
+.. note::
+
+    Here we choose a sphere of radius 22 Å around atom 3 in the first molecule.
```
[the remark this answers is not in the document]
lohedges: My brain is zero indexed, apparently. I'll update this.

doc/source/tutorial/part08/04_diels_alder.rst
lohedges: That's a good idea. I'll add this in as it will make the tutorial a lot cleaner.

chryswoods (dismissed):
This is all excellent. I really like the new tutorial. Very worthy addition for the 2024.3.0 release :-)

chryswoods:
Also just to say the whole code feels really coherent. This is a very powerful new bit of functionality

lohedges:
Thanks, I'm looking forward to it having a bit more visibility. I'm very thankful to the current users who have helped find bugs and iron out the rough edges.

doc/source/tutorial/part08/04_diels_alder.rst
lohedges: I've gone for `d.context()` since there's an equivalent `d.integrator()` already that gets the OpenMM integrator.

PR #250, opened 2024-10-28
Fix issue #249

### LINKED ISSUE #249: [BUG] Sire::Mol::Frame::toByteArray initialises QByteArray incorrectly
As discussed [here](https://github.com/OpenBioSim/somd2/issues/60) the `QByterArray` for used to serialise the trajectory frame is instantiated incorrectly, which causes a segmentation fault when the system size (and `nbytes`) is large. The function also incorrectly uses the `.count()` method to determine the number of bytes for `QByteArray` objects. This is used to count the occurrences of a specific byte. Instead, the `.size()` method should be used to return the number of bytes.

lohedges:
This PR closes #249 and OpenBioSim/somd2#60.

### DIFF
--- a/corelib/src/libs/SireBase/pagecache.cpp
+++ b/corelib/src/libs/SireBase/pagecache.cpp
@@ -662,9 +662,28 @@ void CacheData::cleanUpOnExit()
         {
             if (QThread::currentThread() != c.get())
             {
-                if (not c->wait(100))
+                bool is_finished = false;
+
+                // give the thread a maximum of 10 attempts to finish
+                for (int i = 0; i < 10; i++)
                 {
-                    // kill it
+                    // the thread didn't finish
+                    if (not c->wait(100))
+                    {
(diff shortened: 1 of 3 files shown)


chryswoods (approved):
Thanks for catching this - looks great :-)

lohedges:
No problem. It's one of those where you wonder how it ever worked in the first place. It's also quite surprising (and dangerous) that `QByteArray` appears to be one of the few Qt data structures that doesn't have an aliased `.count()` method, i.e. as an alias for `.size()`. I find it so easy to use `.count()` by mistake since I'm used to using it for `QVector` and `QList`, etc. Perhaps we should be careful to use `.size()` throughout in future.

I've just pushed a minor update to increase the wait period before terminating `QThread` objects when cleaning the `PageCache`. On my machine, this stops the random `FATAL: exception not rethrown` seen on interpreter exit, although it might not be a 100% reliable solution on all platforms. (I also didn't see any crashes if I just commented this part of the code, i.e. didn't attempt to wait for threads to stop at all.)

PR #254, opened 2024-10-29
Fix issue #253

### LINKED ISSUE #253: [BUG] dynamic.minimise() is missing timeout kwarg
I've forgotten to expose the `timeout` kwarg for `dynamics.minimise()`. This _is_ exposed for `DynamicsData`, and the `minimisation` function itself.

### LINKED ISSUE #255: [BUG] Inconsistent behaviour of dynamics.minimise and system.minimisation
I have been trying to track down hangs in the Sire dynamics minimiser and have discovered that the only occur when using `dynamics.minimise()` rather than `system.minimisation()` directly. I assume that they are using the same options behind the scenes, but maybe some kwargs aren't being passed through correctly. The main issue is that I would like to be able to set up a context once (via) and dynamics object, then minimise and run. The hangs require that I use a two-step protocol, where I first create a context to minimise (via `system.minimisation())`, then commit the changes and create a dynamics object from that.

Here is some code to reproduce:

```python
import sire as sr

mols = sr.load_test_files("merged_molecule.s3")
mols = sr.morph.link_to_reference(mols)

map = {
    "check_for_h_by_max_mass": True,
    "check_for_h_by_mass": False,
    "check_for_h_by_element": False,
    "check_for_h_by_ambertype": False,
    "fix_perturbable_zero_sigmas": True,
}

d = mols.dynamics(
    integrator="langevin_middle",
    temperature="300K",
    pressure="1atm",
    timestep="2fs",
    constraint="h_bonds",
    perturbable_constraint="h_bonds_not_heavy_perturbed",
    cutoff="10A",
    cutoff_type="rf",
    dynamic_constraints=True,
    include_constrained_energies=True,
    platform="cuda",
    map=map,
)

print("d.minimise()")
d.minimise()

minimiser = mols.minimisation(
    constraint="h_bonds",
    perturbable_constraint="h_bonds_not_heavy_perturbed",
    cutoff="10A",
    cutoff_type="rf",
    dynamic_constraints=True,
    include_constrained_energies=False,
    platform="cuda",
    map=map,
)
print("minimiser.run()")
minimiser.run()
```

Here I _do_ include constrained energies when setting up the dynamics object, but don't when calling `mols.minimisation()`. All other settings are the same. This gives:

```
d.minimise()
Minimising 1.8: 498 : -48917.7 kcal mol-1                0.4 s 1131.8 steps / s
minimiser.run()
Minimising 1.6: 570 : -49001.2 kcal mol-1                0.5 s 1177.7 steps / s
```

Now if I set the same `include_constrained_energies=False` when instantiating the dynamics object, then `d.minimise()` gets stuck in an infinite minimisation loop. For example, after a few minutes I see:

```
d.minimise()
Minimising 1.0: 79 : -85849.7 kcal mol-1      0%       211 s    0.4 steps / s
```

The minimiser jumps back and forth between a few values and occasionally spikes to a very large value.

The behaviour for these two approaches to minimisation should be identical. I'll try to figure out why they aren't. Given the issues, I wonder if we always want to minimise with the constrained energies included?

lohedges:
This PR closes #253 by exposing the `timeout` kwarg in `dynamics.minimise()`.

Also closes #255 by adding the `include_constrained_energies` kwarg to the map and standardising minimisation settings across all API functions.

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -19,6 +19,8 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 * Fixed instantiaton of ``QByteArray`` in ``Sire::Mol::Frame::toByteArray`` and count bytes with ``QByteArray::size``.
 * Increase timeout before terminating ``QThread`` objects during ``PageCache`` cleanup.
 * Expose missing ``timeout`` kwarg in :meth:`dynamics.minimise()` method.
+* Expose missing ``include_constrained_energies`` kwarg in minimisation function.
+* Make minimisation function settings consistent across API.
 
 `2024.3.0 <https://github.com/openbiosim/sire/compare/2024.2.0...2024.3.0>`__ - October 2024
 --------------------------------------------------------------------------------------------
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -18,6 +18,7 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 * Please add an item to this CHANGELOG for any new features or bug fixes when creating a PR.
 * Fixed instantiaton of ``QByteArray`` in ``Sire::Mol::Frame::toByteArray`` and count bytes with ``QByteArray::size``.
 * Increase timeout before terminating ``QThread`` objects during ``PageCache`` cleanup.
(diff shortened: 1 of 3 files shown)


mb2055 (approved):
Looks good :+1:

PR #258, opened 2024-11-01
Bugfix Boresch Correction

### LINKED ISSUE #248: Definition of force constants inconsistent between Boresch restraints correction and applied restraints
**Describe the bug**

This bug was my mistake and I'll fix it once we agree on the best approach.

The issue is an inconsistency between the definitions of the force constants in the [Boresch restraints Python class](https://github.com/OpenBioSim/sire/blob/4fcc147db1afa4ff5c8e868484eaa18eda3d4ff2/src/sire/restraints/_restraints.py#L21)/ [standard state correction](https://github.com/OpenBioSim/sire/blob/4fcc147db1afa4ff5c8e868484eaa18eda3d4ff2/src/sire/restraints/_standard_state_correction.py#L62), where I've assumed that the definition is consistent with `E = 0.5 * k * x**2`, and [the implementation](https://github.com/OpenBioSim/sire/blob/4fcc147db1afa4ff5c8e868484eaa18eda3d4ff2/wrapper/Convert/SireOpenMM/sire_to_openmm_system.cpp#L97), where the `k`s apply as `E = k * x**2`. The definition as `E = k * x**2` is consistent with other restraints implemented in Sire e.g. [here](https://github.com/OpenBioSim/sire/blob/4fcc147db1afa4ff5c8e868484eaa18eda3d4ff2/wrapper/Convert/SireOpenMM/sire_to_openmm_system.cpp#L208).

One fix would be simply to fix the standard state correction I implemented and keep the current definition consistent with  `E = k * x**2`. However, in this case the `k`s are no longer force constants (because `F = 2kx`), so I should be careful to emphasise this in the documentation. The downside here is that most Boresch restraints are specified in the literature according to the `E = 0.5 * k * x**2` definition, so we'd have to be careful to make this clear. The plus is that the force constants would be consistent with SOMD1.

Alternatively, the OpenMM restraints could be created using the `E = 0.5 * k * x**2` definition. Then the `k`s would correspond to force constants and the Boresch restraints could be supplied in a way which is consistent with most of the literature (but not SOMD1). However, the implementations of other restraints should be changed to be consistent with this, which could be a pain for people currently using the other restraints.

Please let me know what would be best, and I'll make the changes.

Thanks.

fjclark:
This pull request closes #248. I've updated the Boresch restraint correction to be consistent with the definitions of the restraint energies and forces as `E = k * x **2` and `F = 2 * k * x`. To make this clear to the user, I've also updated the restraints docstrings and tutorial page to make this explicit, and have referred to the "k"s as "half force constants". This isn't the cleanest vocabulary, so please let me know if there's a better way to explain this. 

## Suggested reviewers:
@chryswoods, @lohedges

Thanks.

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -16,6 +16,9 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 ---------------------------------------------------------------------------------------------
 
 * Please add an item to this CHANGELOG for any new features or bug fixes when creating a PR.
+* Fixed :func:`sire.restraints.get_standard_state_correction` to be consistent with the definition of
+  the "force constanst" as ``F = 2 ** k ** x`` (rather than ``F = k ** x``). Updated docstrings and
+  restraints documentation to make it clear how the force constants are defined.
 * Fixed instantiaton of ``QByteArray`` in ``Sire::Mol::Frame::toByteArray`` and count bytes with ``QByteArray::size``.
 * Increase timeout before terminating ``QThread`` objects during ``PageCache`` cleanup.
 * Expose missing ``timeout`` kwarg in :meth:`dynamics.minimise()` method.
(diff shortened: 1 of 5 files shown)


lohedges (approved):
Thanks for this. I'm happy with the changes.

PR #273, opened 2025-01-15
Add support for replica exchange and REST2

### LINKED ISSUE #256: [BUG] Frames and energies are always saved after dynamics.run()
I am using Sire dynamics for replica exchange. In doing so, I need to run short cycles at the `energy_frequency` then swap replicas. However, if I set a `frame_frequency` to larger than my simulation run time I find that a frame is still saved after `dynamics.run()` completes. In fact, it seems that energies and frames are _always_ saved when it exits. This is _massively_ impacting performance, since I am effectively saving frames at the (very short) `energy_frequency`.

As a solution, the code should be updated so that calls to `.run()` are not independent of each other, i.e. it remembers how many steps were performed before and calculates the time till saving the energy or a frame accordingly.

### LINKED ISSUE #270: Sire::IO::SDF can't handle a stereo value of 3
Hello!
I am just starting to work with BioSimSpace and got a problem with reading one of my SDF files.
I run the simplest  code:
```
import BioSimSpace as BSS
lig = BSS.IO.readMolecules('ligand.sdf')[0]
```

I consistently get the following error:
```
INFO:numexpr.utils:Note: NumExpr detected 64 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 16.
INFO:numexpr.utils:NumExpr defaulting to 16 threads.
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│a3fe/test/script.py:2   │
│ in <module>                                                                                      │
│                                                                                                  │
│   1 import BioSimSpace as BSS                                                                    │
│ ❱ 2 lig = BSS.IO.readMolecules('ligand.sdf')[0]                                                  │
│   3 #lig = BSS.IO.readMolecules("ligand_benzene.sdf")[0]                                         │
│   4 #lig_param = BSS.Parameters.openff_unconstrained_2_0_0(lig).getMolecule()                    │
│   5 #BSS.IO.saveMolecules("lig_param", lig_param, ["prm7", "rst7"])                              │
│                                                                                                  │
│ Miniconda3/envs/a3fe/lib/python3.12/site-packages/BioSimSpace/I │
│ O/_io.py:556 in readMolecules                                                                    │
│                                                                                                  │
│    553 │   │   │   │   if _isVerbose():                                                          │
│    554 │   │   │   │   │   raise IOError(msg) from e0                                            │
│    555 │   │   │   │   else:                                                                     │
│ ❱  556 │   │   │   │   │   raise IOError(msg) from None                                          │
│    557 │                                                                                         │
│    558 │   # Add a file format shared property.                                                  │
│    559 │   prop = property_map.get("fileformat", "fileformat")                                   │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯

```
While when I try to read a benzene sdf file (which was obtained the same way as the ligand.sdf, RDKit version 2024.03.2) everything works just fine.

I have attached my input files: ligand.sdf (where the error occurs) and ligand_benzene.sdf (works fine).
[ligands.zip](https://github.com/user-attachments/files/18201091/ligands.zip)

OS: Rocky Linux 8.9 (Green Obsidian)
Version of Python: 3.12.8
Version of BioSimSpace: 2024.4.0.dev+34.g357beaf2

Could you help me please?
Thank you!

### LINKED ISSUE #274: [BUG] Positive formal_charge isn't converted to SDF charge field correctly
Round-trip read/write of SDF files is failing due to incorrect handling for positive formal charge values on write. According to this resouce we should have:

| Formal charge | SDF value |
|--------------------|------------------|
| -3 |  7 |
| -2 | 6 |
| -1 | 5 |
| 0 | 0 |
| 1 | 3 |
| 2 | 2 |
| 3 | 1 |

The code is handled here https://github.com/OpenBioSim/sire/blob/ecfef72873f56c6debc0696aaaf1b771c22767e6/corelib/src/libs/SireIO/sdf.cpp#L953-L976

This has no conditional for positive formal charge, so it just takes the formal charge value, i.e. 1 would be written as 1, not 3. I'm also not sure what one does about the additional "doublet radical" option, which should have a value of 4?

lohedges:
This PR adds the low-level changes to `sire` required to support hamiltonian replica exchange and REST2 simulations using `somd2`. The PR modifies some existing `sire` functionality, adds some new functionality, and provides several fixes and enhancements.

_Modified functionality_:

* In order to facilitate replica exchange (repex) it was necessary to modify `dynamics.run()` such that samples (trajectory and energy frames) were not always recorded on exit. When performing repex we use `dynamics.run()` to perform many short cycles, only saving energies after each. Performance was massively impacted by the existing code where trajectory frames are always saved. The change closes #256, with the sampling frequencies for energies and frames now decoupled from the run method, i.e. the are only ever recorded at a multiple of the frequency. This is validated in [this](https://github.com/OpenBioSim/sire/blob/3eff41b7982683ccc0bed50fb0dce18bfc32a2e7/tests/mol/test_dynamics.py#L84) test. Note that this doesn't affect the ability to run a short dynamics plot and commit changes to the system, since the coordinates are updated from the current OpenMM state when `.commit()` is called, so is unrelated to trajectory sampling, etc.

* The handling of `NaN` errors has been tweaked so that it's possible to catch and fix via minimisation when an exception is raised on any step within the `dynamics.run()`, rather than just on the first step, as long as no sampling has been performed (energy or trajectory frames). This improves the robustness of the dynamics blocks within `somd2` and allows us to make use of `auto_fix_minimise=True` for all simulation types. Previously we ran things within our own `try/except` block with custom error handling.

_New functionality_:

* REST2 is supported by the new `rest2_scale` and `rest2_selection` kwargs to the `.dynamics` method. The first defines the scale factor to use for REST2 simulations, which specifies the _temperature_ of the REST2 region relative to the rest of the system. I chose temperature, since this is what is commonly reported in the literature, although some codes use beta instead. I'm not sure if it's easier to think about making the REST2 region _hotter_ or the interactions _softer_. I'm happy to use the inverse if preferred. (Internally the scale factor is inverted before scaling force constants, etc.) The `rest2_selection` is used to specify additional atoms to include in the REST2 region via a `sire` selection string. By default, all atoms in perturbable molecules are included. The selection provides a way of specifying other relevant atoms, e.g. ones in the protein binding site. When perturbable molecules are specified, then _only_ those perturbable atoms are placed in the REST2 region, i.e. atoms that aren't specified are unaffected. This provides a way of selecting a sub-region of a perturbable molecule if desired.

* Modifications to dihedral and nonbonded terms are handled in two ways. Firstly we deal with terms in the perturbable region via the lambda lever code. To do so, we create a mask for atoms in the perturbable molecule that are affected by the REST2 modifications, which is determined via the selection described above. Secondly, we need to keep track of which torsion terms are proper dihedrals, since impropers are not modified. With this in place, we can then apply the REST2 scaling to the required terms when `LambdaLever::setLambda()` is called. Terms corresponding to atoms that are not within perturbable molecules are handled separately using the OpenMM Python API via the approach used by [meld](https://github.com/maccallumlab/meld/blob/master/meld/runner/transform/rest2.py). The terms that are modified are pre-computed when instantiating the `SOMMContext` and are modified when calling `SOMMContext.set_lambda()` when a `rest2_selection` containing non-perturbable atoms has been specified. There is a fairly extensive [unit test](https://github.com/OpenBioSim/sire/blob/feature_repex/tests/convert/test_openmm_rest2.py) to make sure the appropriate terms are scaled correctly. This covers: 1) A perturbable molecule; 2) A selection of atoms within a perturbable molecule, and 3) A non-perturbable molecule.

* When calling `dynamics.run()` a user can now pass an optional `rest2_scale_factors` kwarg. This specifies the REST2 scale factor to use at each of the `lambda_windows` that are specified for energy evaluation. We only validate that there are the correct number of entries, not that the scale factors are valid, e.g. they should be 1 at the end states, etc. This validation is done at the `somd2` level.

_Fixes_:

* I've fixed an issue with the Sire OpenMM minimiser which meant that exceptions where thrown _prior_ to the thread state being restored, which would cause the program to crash. This closes OpenBioSim/somd2#63.

* I've added a small update to the `sire.qm` code to search for the `QMForce` by name before returning. (It is always named `QMForce`.) Currently it is always the first force that is added to the system so we can search by index. However, I'd rather make this change now so that we don't run into any surprises if the ordering changes in future. (This change is still covered by an existing unit test.)

* Closes #270 by adding handling for double bond stereo info from SDF files.

* Closes #274 by adding handling for positive formal charge on SDF write.

## Checklist

@mb2055: Adding you since you are currently running simulations so can comment on any issues you encounter.
@chryswoods: Adding you so that you are aware of changes to the existing functionality. Let me know if any of these changes impact anything in ways that I might not have anticipated.


lohedges:
Looks like there are some veryslow tests that are failing. (Forgot to run these locally.) Will debug and push a fix.

lohedges:
Now fixed.

PR #275, opened 2025-01-20
Update to the Alchemical Dynamics docs

mb2055:
A small update to tutorial 6, part 2 - Alchemical Dynamics. Makes the plots a bit clearer and ensures that they show the actual output of the dataframes in question. Also fixes a few spelling mistakes and a line of code that used the incorrect call.

Tagging in @lohedges so that this can be added to the `feature_repex` PR.
Also tagging @akalpokas in case I missed anything or there are more changes to be made.


lohedges (approved):
Thanks, @mb2055, this looks good to me.

PR #277, opened 2025-01-24
Exclude alchemical ions from the REST2 region

lohedges:
This PR applies a small fix by excluding alchemical ions from the REST2 region. These are tagged in `somd2` using the `is_alchemical_ion` property, then excluded via a REST2 mask.

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -37,6 +37,8 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 
 * Added support for angle and dihedral restraints which can be used in alchemical and standard simulations.
 
+* Allow user to compute energy trajectory over a subset of the lambda windows for each lambda.
+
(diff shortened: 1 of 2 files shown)


mb2055 (changes requested):
There seems to be an issue when running without explicitly defining `lambda_windows`.
To recreate:
```python
import sire as sr

mols = sr.load(sr.expand(sr.tutorial_url, "merged_molecule.s3"))

mols = sr.morph.link_to_reference(mols)

mols = mols.minimisation(lambda_value=0.5).run().commit()

d = mols.dynamics(lambda_value=0.5, timestep="4fs", temperature="25oC")

d.run("10ps", save_frequency="1ps")
```
raises `TypeError: object of type 'NoneType' has no len()` due to line 393. Can probably be fixed by just setting `lambda_windows` equal to `lambda_value` in the case where the user doesn't set it.

lohedges:
Ah yes, good point, I'll just only store `num_lambda_windows` if not `None`.

mb2055 (approved):
Looks good

PR #290, opened 2025-02-11
Fix issue #287

### LINKED ISSUE #287: [BUG] Allow user to force RDKit stereo inference
In some cases SDF files might be incorrect, so we need to expose an option to allow the user to force the use of the manual inference routine. This can handled via a map option with a kwarg exposed via BioSimSpace.

### LINKED ISSUE #289: [BUG] SDF reader sets incorrect formal charge
The SDF reader sets the incorrect formal charge on atoms when it is a positive value in the SDF file. This has been fixed for write (see [this](https://github.com/OpenBioSim/sire/issues/274) issue), but I forgot to actually set the correct charge on read. This means that all positive formal charge for atoms coming from SDF has been wrong since inception. Previously we would get a correct round-trip read/write since the value was unaffected, but the loaded molecule would have the wrong formal charge property. Now we don't get read/write self-consistency since the atomic formal charge is set correctly, then correctly converted on write, but to a different value.

lohedges:
This PR closes #287 by adding a map option to force the fresh inference of stereochemistry when converting to RDKit format. Previously there was a bug which meant that the code _always_ took this pathway, meaning that it didn't always preserve what was in the original SDF file. This new approach allows the user to take _both_ approaches, then compare the result in RDKit, i.e. they can see if the information in the file agrees with what the inference suggests.

The PR also closes #289 where the `formal_charge` property of an atom was set incorrectly for certain positive charges. This is because the value in the SDF file isn't the actual charge, e.g:

| Formal charge | SDF value |
|--------------------|------------------|
| -3 |  7 |
| -2 | 6 |
| -1 | 5 |
| 0 | 0 |
| 1 | 3 |
| 2 | 2 |
| 3 | 1 |

For example, a value of 3 in the file meant that an atom had +3 charge, not +1. This was ultimately the source of several BioSimSpace parameterisation issues, e.g. like the one posted [here](https://github.com/OpenBioSim/biosimspace/issues/370). While I had correctly fixed the conversion of the formal charge property _back_ to the SDF field, I hadn't noticed the same bug was present when setting the property in the first place. This means that all parameterisation starting from SDF since BioSimSpace 2024.3.0 will have been incorrect if there was an atom with +1 charge (SDF value 3). As such, it might be worth testing some know systems to see if anything has changed.

The fix to the formal charge has meant that we can re-enable an XFAILED RDKIt SMARTS inference test that was disabled following the previous edits. Those meant that it no longer passed, but the formal charge was the underlying culprit.

### DIFF
--- a/corelib/src/libs/SireIO/sdf.cpp
+++ b/corelib/src/libs/SireIO/sdf.cpp
@@ -235,11 +235,11 @@ namespace SireIO
                 case 0:
                     return 0;
                 case 1:
-                    return 1;
+                    return 3;
                 case 2:
                     return 2;
                 case 3:
-                    return 3;
+                    return 1;
                 case 4:
                     return 0;
                 case 5:
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -17,6 +17,8 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 ------------------------------------------------------------------------------------------
 
 * Please add an item to this CHANGELOG for any new features or bug fixes when creating a PR.
+* Allow user to force fresh inference of stereochemistry when converting to RDKit format.
+* Fix setting of positive formal charge when reading SDF files.
 
 `2024.4.0 <https://github.com/openbiosim/sire/compare/2024.3.1...2024.4.0>`__ - Feb 2025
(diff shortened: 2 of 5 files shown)


mb2055 (approved):
All good :+1:

PR #299, opened 2025-02-17
Fix issues #296, #297, and #298

### LINKED ISSUE #296: [BUG] Sire-to-RDKit doesn't preserve SDF tags
The Sire-to-RDKit conversion doesn't preserve any SDF tags that have been loaded into the molecule, which are present in the `sdf_data` molecular property. These shoudld be added as properties of the RDKit molecule so that they can be preserved when writing to SDF via RDKit.

### LINKED ISSUE #297: [BUG] SDF molecule name should be given to Sire molecule
When an SDF file contains a molecule name, then the Sire molecule should take that name on read.

### LINKED ISSUE #298: [BUG] AmberRst7 parser fails if there is a blank line at the end of file
The `SireIO::AmberRst7` parser fails if there is a blank line at the end. The line can contain valid data, which it is trying to parse. We should check for an empty line and skip.

lohedges:
This PR closes #296 by setting and SDF tags as properties of the RDKit molecule during conversion. (The SDF metada is stored internally in Sire as a `sdf_data` property on molecule.) On conversion back to Sire, all non-internal properties of a RDKit molecule are now stored as an `rdkit_data` property. We can't write back to `sdf_data`, since the RDKit molecule doesn't tag where the properties come from, e.g. were they set when loading from SDF, or otherwise, so there is no clean separation of what data came from where. A user _could_ allow these tags to be used when writing to SDF by using the property map, e.g. by remapping `sdf_data` to `rdkit_data`. In any case, the main use case here is the ability to convert an SDF molecule to RDKit, then later write to SDF using RDKit and retain all of the tags in the original file, which this does.

The PR also closes #297 by setting the `MolName` of any molecule loaded from SDF to the name in that file. Currently it takes a default molecule name of `MOL`. There is also a `name` molecular property that is set, but CI failures made me realise that this can also be used for other information, e..g. the `neopentane-methane` merged system appeared to have some kind of atom selection as the `name` property.

```python
In [1]: import sire as sr

In [2]: mols = sr.load_test_files("neo_meth_scratch.bss")


In [3]: mols[0].property("name0")
Out[3]:
SireMol::AtomStringProperty( size=17
0: C1
1: C2
2: C3
3: C4
4: C5
...
12: H13
13: H14
14: H15
15: H16
16: H17
)
```

The PR also closes #298 by adding special handling for velocity records when the number of atoms in an AMBER RST7 file is 2 or less. If so, then the velocity line could be blank if no data is present. (There is a check to see how many lines should be exist if velocities are present, which this passes, but the line contains no data.) I'm not sure what parser is adding a blank line at the end, but it is part of the Exs pipeline.

### DIFF
--- a/corelib/src/libs/SireIO/amberrst7.cpp
+++ b/corelib/src/libs/SireIO/amberrst7.cpp
@@ -357,6 +357,15 @@ void AmberRst7::parse(const PropertyMap &map)
         // we have already validated that there are enough lines
         const QString &line = l[linenum];
 
+        if (natoms <= 2 and line.trimmed().isEmpty())
+        {
+            // if there are only two atoms, then the final line might not be reserved
+            // for velocities
+            vels.clear();
+
+            return;
+        }
+
         if (line.length() < column + 36)
         {
             errors.append(QObject::tr("Cannot read the velocities for the atom at index %1 as "
--- a/corelib/src/libs/SireIO/sdf.cpp
+++ b/corelib/src/libs/SireIO/sdf.cpp
@@ -1898,6 +1898,9 @@ MolEditor SDF::getMolecule(int imol, const PropertyMap &map) const
         mol.setProperty(map["sdf_counts"], SireBase::wrap(sdf_counts));
     }
 
+    // set the name of the molecule
+    mol.rename(MolName(sdfmol.name));
+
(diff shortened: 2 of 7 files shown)


chryswoods (approved):
Look good :-)

PR #306, opened 2025-03-09
Fix InvalidMatchSpec error when installing packages using setup.py.

CHENFENG1020:
No published issue related to this.

In `setup.py`, within the `conda_install` function, `dep = f'"{dependency}"'` leads to an conda error when installing a package with requied version, such as `gemmi>=0.6.4`:  `InvalidMatchSpec: Invalid spec '"gemmi>=0.6.4"': Invalid version '0.6.4"': invalid character(s)`   (*conda version:  25.1.1)
  
To Reproduce: Create a new env and run `python setup.py install_requires`, or simply: `'conda' 'install' '"gemmi>=0.6.4"'` (How setup.py install package).
  
The setup.py print:`Installing packages using: /xxx/bin/conda install "gemmi>=0.6.4"`, but it actually using:`'/xxx/bin/conda' 'install' '"gemmi>=0.6.4"'`

## Suggested reviewers:
@chryswoods, @lohedges

Also updated the installation command in the documentation.

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -25,6 +25,7 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 * Handle missing velocity data for ``AMBER`` RST7 files with only a few atoms.
 * Preserve SDF metadata when converting to ``RDKIt`` format.
 * Fixed unbound local variable when simulation lambda value is not in ``lambda_windows`` array.
+* Fix InvalidMatchSpec error when installing packages using setup.py.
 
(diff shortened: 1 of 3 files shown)


lohedges:
Many thanks. I've resolved this locally but hadn't pushed a fix (mostly because it doesn't affect the Python versions used for the CI).

@chryswoods: I think the fix on your feature branch in slightly different, albeit with the same affect. Are you okay with this, or should I patch in your change instead. (Mostly to avoid conflicts later.)

PR #321, opened 2025-06-24
Fix #319 take 2

### LINKED ISSUE #319: [BUG] Energy trajectory issues
In rare circumstances it appears that the energy trajectory can get corrupted somehow, with missing entries and incorrect and inconsistent time intervals.

The current line of thinking is that this could be related to recovering from unstable simulations triggered by high kinetic energies or NaNs in OpenMM. We have specific handling to catch this, re-minimise, then attempt to continue the simulation. If the simulation clock isn't reset correctly then it's possible that incorrect time records would appear in the energy trajectory, or that existing values could be overwritten.

lohedges:
This PR closes #319 by additionally resetting the internal clock when an OpenMM NaN crash is caught. Previously only the step counter was reset, but this led to incorrect time stamps for subsequent energy and trajectory frames since the clock was reset to zero, rather than the start time of the last dynamics block.

I've also updated the energy trajectory code to store keys in rounded form to avoid issues from numerical drift when times are updating over a long trajectory. The code now outputs a warning whenever keys are overwritten.

### DIFF
--- a/corelib/src/libs/SireMaths/energytrajectory.cpp
+++ b/corelib/src/libs/SireMaths/energytrajectory.cpp
@@ -28,6 +28,8 @@
 
 #include "energytrajectory.h"
 
+#include "SireBase/console.h"
+
 #include "SireUnits/units.h"
 
 #include "SireID/index.h"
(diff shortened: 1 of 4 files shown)


lohedges:
I've removed the rounding since it doesn't work reliably when the energy frequency is very small. This isn't likely an issue in practice, but is causing problems with the unit tests.

lohedges:
Looks like it's just complaining about the timestamp, so we should ignore that line.

lohedges:
Now fixed. I've just changed the test to skip the first line, which is where the timestamp is found.

PR #333, opened 2025-07-17
Fix issue #332

### LINKED ISSUE #332: [BUG] evaluate_xml_force incorrectly using IntereactionGroups
The XML parser used to evaluate custom forces is incorrectly using interaction groups. While they are parsed correctly, there is a double loop over the first set, rather than both sets. This leads to incorrect energies.

lohedges:
This PR closes #332 by correctly looping over interactio groups when evaluating energies from custom forces in an OpenMM XML file.


mb2055 (approved):
Just a minor spelling mistake, otherwise all good :+1:

PR #336, opened 2025-07-23
Fix issue #335 and #337

### LINKED ISSUE #335: Remove RDKit down-constraint from sire conda package
We recently found that the root cause of a [user report](https://github.com/openforcefield/openff-toolkit/issues/2089#issuecomment-3105292753) was due to the sire 2025.1.0 conda package down-constraining rdkit (constraint `rdkit >=2024.3.5,<2024.3.6.0a0` which caused an error due to a change in some sanitization behavior that OpenFF NAGL relies on). Would it be possible to remove this down-constraint from the sire package?

### LINKED ISSUE #337: [BUG] librdkit-dev doesn't enforce boost constraint
As recommended by the `rdkit` conda package maintainer, we now require only `librdkit-dev` as a host requirement and enforce a runtime pin on `rdkit` via the version of `librdkit-dev`. However, it turns out that only the `rdkit` package depends on `boost`, so we can build against a version of `libboost` that is incompatible with the run time pin. For example, `librdkit-dev-2025.4.3` might be resolved in the host requirements and we also resolve `libboost-1.8.2`. The runtime pin for `rdkit` is then `2025.4.3` to `2025.4.4`, but this requires `libboost-1.8.6`. As such, we need to add `rdkit` to the host requirements too.

lohedges:
This PR closes #335 and closes #337 by switching to the new Boost package naming scheme. We now require `libboost-devel` and `libboost-python-devel` as host requirements. The change means that we are now compatible with newer versions of RDKit, which have also been built against the new Boost packages.

### DIFF
--- a/corelib/CMakeLists.txt
+++ b/corelib/CMakeLists.txt
@@ -370,10 +370,17 @@ else()
   CHECK_CXX_COMPILER_FLAG( "-std=c++17" SIRE_HAS_CPP_17 )
 endif()
 
+if ( CMAKE_CXX_COMPILER_ID MATCHES "MSVC" )
+  CHECK_CXX_COMPILER_FLAG( "/std:c++20" SIRE_HAS_CPP_20 )
+else()
+  CHECK_CXX_COMPILER_FLAG( "-std=c++20" SIRE_HAS_CPP_20 )
+endif()
+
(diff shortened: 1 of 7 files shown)


lohedges:
Oh wonderful, this now fails with [this](https://github.com/jtv/libpqxx/issues/434) issue.

PR #347, opened 2025-08-13
Fix issue #346

### LINKED ISSUE #346: [BUG] Velocity data can overflow fixed precision format
Due to the fixed-precision format of many coordinate files, large velocities can easily overflow the precision causing an exception to be triggered on write. In this case it would be good to provide an option to `sr.save` to allow the user to skip the writing of velocity data if desired, i.e. only positions will be written. (Obviously a binary format could be used, but a user may wish to use a human readable format.)

A trivial solution is to remap the `velocity` property to a null value using the map, e.g.:

```python
sr.save(mols, "test.rst7", map={"velocity": "cat"}
```

This could be added to the map automatically if a specific kwarg is passed, e.g.:

```python
sr.save(mols, "test,prm7", save_velocities=False)
```

lohedges:
This PR closes #346 by allowing the user to control whether velocities are written to the output files. I appreciate that this could also class as a feature, but feel that there should be a way to control this since this commonly happens during equilibration pipelines for our FEP setup. (The exception should definitely be raised, but then there should be an easy way for the user to control whether or not velocities are needed.)

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -33,6 +33,10 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 
 * Added a feature for applying RMSD restraints using the OpenMM CustomCVForce/RMSDForce functionality.
 
+* Added a ``save_velocities`` kwarg to :func:`sire.save()` to allow a user to control whether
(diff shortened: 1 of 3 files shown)


chryswoods (approved):
Very neat fix, and solves a pain point for people.

PR #348, opened 2025-08-13
Fix issue #344

### LINKED ISSUE #344: [BUG] Sire doesn't save periodic box vectors to PDB
PDBs support periodic boxes via the `CRYST1` record. However, when using sire.save() to write a system to a .pdb file, no periodic box vectors are written even when the `space` property is set. This results in information being lost when saving NPT simulation results to a PDB

lohedges:
This PR closes #344 by adding support for parsing `CRYST1` records in PDB files. Depending on the box angles, these are converted to a `PeriodicBox` or `TriclinicBox` object, then stored as the `space` property of the system. On write, the `space` property is converted back to an appropriate `CRYST1` record.

A unit test checks for round-trip self-consistency for a standard Cresset tricnlic box. The test assumes 2dp accuracy in the box vectors, since the PDB format stores the `CRYST1` angles to 2dp.

### DIFF
--- a/corelib/src/libs/SireIO/pdb2.cpp
+++ b/corelib/src/libs/SireIO/pdb2.cpp
@@ -49,6 +49,9 @@
 
 #include "SireUnits/units.h"
 
+#include "SireVol/periodicbox.h"
+#include "SireVol/triclinicbox.h"
+
 #include <QFile>
 #include <QtMath>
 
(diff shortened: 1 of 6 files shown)


tom-potter-cresset (dismissed):

chryswoods (approved):
Looks good - it's good to support this, but we have to be careful that it is "non-standard" to add periodic box data to the crystal section. This section is more properly used to describe the crystal (I think?) and so we may need to be careful not to add periodic boxes to structures where cryst is not used for this. But, PDB is very non-standard, so I'd hope this wouldn't be a problem for most use cases.

In general, periodic box info from any other file format should take precedence over periodic box info from PDB.

PR #359, opened 2025-09-17
Fix issue #358

### LINKED ISSUE #358: [BUG] OPC not detected as a water topology by GroTop parser
I must have forgotten to update the `GroMolType::isWater` method for OPC water, so the new topology currently isn't being detected as water. This means that `[ settles ]` records are missing from the GROMACS topology file, so constrained water simulations aren't currently possible.

lohedges:
This PR closes #358 by fixing the handling for detecting writing 4- and 5-point water models in the `GroTop` parser. Previously, despite stating otherwise, the code only correctly handled 3-point models and incorrectly assumed that all water models to be written would be AMBER-style, i.e. including explicit HH _and_ OH bonds in the force field, which would be used to determine the values to use in the `[ settles ]` section. For GROMACS, there is no explicit OH bond in any model, so this logic breaks down, meaning the default TIP3P parameters would be used for _all_ models. The fix also includes the manual addition of the correct `[ virtual_sites3 ]` records for each model, which are taken from the GROMACS water templates that are included with a GROMACS install and used for solvation with BioSimSpace. A unit test confirms that the correct records are written for each water model.

A quick summary:

* 4- and 5-point waters are now detected as water topologies in the parser.
* Correct settles records are added for 4- and 5-point waters.
* Exclusions are correctly written for 4- and 5-point waters.
* Virtual site records are now written for 4- and 5-point waters.

In addition to this, I'll check some GROMACS simulations using these water models via BioSimSpace and will update if there are any issues.

### DIFF
--- a/corelib/src/libs/SireIO/grotop.cpp
+++ b/corelib/src/libs/SireIO/grotop.cpp
@@ -2610,42 +2610,64 @@ QStringList GroMolType::settlesLines(bool is_lambda1) const
 
     QStringList lines;
 
+    // lambda function to check whether a four point water model
+    // is OPC water, which is determined by the virtual site charge
+    // value being < -1.1
+    auto is_opc = [this, is_lambda1]() -> bool {
+        if (is_lambda1)
+        {
+            for (const auto &atm : atms1)
+            {
+                if (atm.mass().value() < 1.0) // virtual site
+                {
+                    if (atm.charge().value() < -1.1)
+                        return true;
+                    else
+                        return false;
+                }
+            }
+        }
+        else
+        {
+            for (const auto &atm : atms0)
+            {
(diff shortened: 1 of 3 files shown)


lohedges:
Just a note regarding the unit test. The logic here is to validate the detection of the correct water topology within the `GroTop` parser on write. Even though some (but not all) of the records may be present in the templates that are read, they are not stored in any molecular or system property, so need to be correctly reconstructed on write.

lohedges:
Looks like the Windows build has hung. (Not that I'm expecting it to pass.) Also note that the Python 3.10 on Linux took about 2 hours more than the other Linux builds, which seems bizarre.

mb2055 (approved):
Looks good :+1:

PR #376, opened 2025-10-28
Add support for dynamics crash reports

lohedges:
This PR adds supports for saving crash reports during Sire dynamics runs using the `save_crash_report` kwarg when calling `dynamics.run()`. If specified, the following files will be saved during a crash:

* A log file with current lambda and energy contribution from each force.
* An XML file containing the OpenMM system.
* A NumPy text file containing the atomic positions.

All files correspond to the system at the _start_ of the dynamics block, not the crash, which might not be available. This allows a user to replay a crash. Files are given a UUID name that is consistent across files, e.g. `crash_UUID.log`, `system_UUID.xml`, and `positions_UUID.txt`. I'm not sure whether we should allow the user to pass a name, or create the UUID for the dynamics object itself, then increment some file index, e.g. if there are multiple crashes along a trajectory. This probably isn't that important, since this is really intended to only be turned on when debugging unstable simulations to capture examples of soft-crashes. Happy to take suggestions, though.

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -31,6 +31,8 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 
 * Add support for ``openmm.MonteCarloMembraneBarostat`` in Sire-to-OpenMM conversion.
 
+* Added support for saving crash reports during Sire dynamics runs.
+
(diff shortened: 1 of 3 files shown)


mb2055 (approved):
All good :+1:

PR #380, opened 2025-11-03
Add inverse bond/distance restraint and fix issue #379

### LINKED ISSUE #379: [BUG] Periodic boundary conditions for OpenMM restraints
At present, the restraint forces that are created during the Sire-to-OpenMM conversion are set to use periodic boundary conditions. For bonded restraints, this doesn't make sense since a restraint could be satisfied in different ways depending on whether it is measured across the boundary or not, e.g. a molecule could extend, or contract, e.g. in a ring-breaking simulation.

I'm not sure whether we want to update all of the restraint classes in `Sire::MM` to add a `use_pbc` option in the constructor, or simply disable periodicity in the conversion layer. The former is more general, but requires a lot of updates and changes to the streaming operator versioning for all restraint classes. One complexity is the "distance" restraint, which is also used for "bond" restraints. I imagine the desired periodicity is different depending on whether you want to restraint a bond, or keep an atom close to some reference position.

lohedges:
This PR adds support for `InverseDistanceRestraint(s)` and `InverseBondRestraint(s)`, which allows two atoms to be kept apart. This is useful for charge change perturbations where it is desirable to keep alchemical ions away from the perturbing ligand.

The PR also closes #379 by allowing the user to specify whether periodic boundary conditions should be used when creating restraints via the new Sire Python API. By default, _positional_ and _distance_ restraints use PBC, and _bonded_ restraints don't. This is needed to avoid issues with bonded restraints being incorrectly satisfied across the periodic boundary, e.g. when a broken bound is being restrained.

I decided to only fix this via the new Python API for simplicity, i.e. by adding a `_use_pbc` attribute to the restraints objects returned by the functions in `sire.restraints`, then passing this through to the Sire-to-OpenMM conversion layer via the map. This avoided the need to update all of the corresponding restraint classes in `Sire::MM`, along with adding new versions for all streaming operators. This could still be done in future, since the Python API wouldn't need to change.

I've added a unit test that confirms that the PBC flag is set correctly for different restraints and that this propagates through to the OpenMM force. In particular, I've tested that the `BondRestraintForce`, which can be used for _distance_ and _bond_ restraints, is set an appropriate default depending on whether it is used as a _distance_ or _bond_ restraint, i.e. distance does use periodicity, bond doesn't. If we strictly want these to be synonyms, then I can standardise the behaviour. My interpretation was that _bond_ refers to physically bonded atoms, whereas _distance_ can refer to any two atoms, although I guess they could still be part of the same molecule, so issues could occur with PBC. (Maybe we could check for this if needed, then set a default based on whether they are in the same molecule?)

This will be the last PR before the 2025.3.0 release. Following this, I'll create a new BioSimSpace release, then initial conda packages for our other tools, i.e. `ghostly`, `loch`, and `somd2`. This will allow everyone to work with a consistent set of packages for the purposes of running initial benchmarks.

@akalpokas: Tagging you for reference.

### DIFF
--- a/corelib/src/libs/SireMM/anglerestraints.cpp
+++ b/corelib/src/libs/SireMM/anglerestraints.cpp
@@ -488,13 +488,13 @@ AngleRestraints AngleRestraints::operator+(const AngleRestraints &restraints) co
 }
 
 /** Set whether or not periodic boundary conditions are to be used */
-void AngleRestraints::setUsesPeriodicBoundaryConditions(bool use_pbc)
+void AngleRestraints::setUsesPbc(bool use_pbc)
 {
     this->use_pbc = use_pbc;
 }
 
 /** Return whether or not periodic boundary conditions are to be used */
-bool AngleRestraints::getUsesPeriodicBoundaryConditions() const
+bool AngleRestraints::usesPbc() const
 {
     return this->use_pbc;
 }
--- a/corelib/src/libs/SireMM/anglerestraints.h
+++ b/corelib/src/libs/SireMM/anglerestraints.h
@@ -161,8 +161,8 @@ namespace SireMM
         AngleRestraints operator+(const AngleRestraint &restraint) const;
         AngleRestraints operator+(const AngleRestraints &restraints) const;
 
-        void setUsesPeriodicBoundaryConditions(bool use_pbc);
-        bool getUsesPeriodicBoundaryConditions() const;
+        void setUsesPbc(bool use_pbc);
+        bool usesPbc() const;
 
     private:
         /** List of restraints */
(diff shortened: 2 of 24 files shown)


lohedges:
Not sure what happened, but it appears no test input files could be read. Everything worked locally before pushing. Guess GiHub might have had a connection issue.

lohedges:
The PR also adds support for specifying the OpenCL platform index when creating an OpenMM context via the `opencl_platform_index` map option, which will be exposed via `somd2`.

@mark-mackey-cresset: Tagging you for reference. Let me know if there are any other platform related options that need exposing, other than those that are already available. (See [here](https://sire.openbiosim.org/cheatsheet/openmm.html#choosing-options) for a list.)

PR #384, opened 2025-11-25
Fix issue #383

### LINKED ISSUE #383: [BUG] Null EMLE OpenMM force causes issues on certain platforms
A null EMLE force is used to add a global parameter to an OpenMM-ML context, allowing interpolation of the EMLE force independent of the OpenMM-ML one. This appears to cause random issues on some platforms since the empty force expression cannot always be parsed. This can be fixed by replacing `""` with `"O"`.

### LINKED ISSUE #385: [BUG] Incorect delta expression for soft-core coulomb term in docs and code
The expression used for `delta` in the soft-core potential [docs](https://sire.openbiosim.org/tutorial/part07/03_ghosts.html#zacharias-softening) is wrong. The expression used should be `detla = shift_coulomb^2 * alpha` as in the comments within the code. The OpenMM expression was updated to be consistent with the website docs, so is also incorrect.

lohedges:
This PR closes #383 by using `"0"` for the null EMLE interpolation force. 

Also closes #385.

@JMorado

### DIFF
--- a/wrapper/Convert/SireOpenMM/sire_to_openmm_system.cpp
+++ b/wrapper/Convert/SireOpenMM/sire_to_openmm_system.cpp
@@ -1359,7 +1359,7 @@ OpenMMMetaData SireOpenMM::sire_to_openmm_system(OpenMM::System &system,
         {
             nb14_expression = QString(
                                   "coul_nrg+lj_nrg;"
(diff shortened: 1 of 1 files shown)


JMorado:
Thanks.

PR #413, opened 2026-03-24
Fix run_constrained key

lohedges:
This PR fixes an invalid conda key that appeared in the switch to `rattler-build`. The key should be `run_constrained` rather than `run_constraints`. Weirdly, the resulting `info.json` files appear to be correct, so perhaps this was an alias. I'll fix for consistency anyway so that we aren't relying on any undocumented behaviour.


lohedges:
Actually, the original key was correct, just the `rattler-build` docs were a bit sparse in this area.

PR #425, opened 2026-04-16
Fix issue #350

### LINKED ISSUE #350: [BUG] Incorrect charges from Sire to RDKit conversion
**Describe the bug**
Converting some molecules from Sire to RDKit results in some strange formal charges and bonded structure. For example, the following molecule:

Image

becomes (formal charges labelled):

Image

This only happens if loading from a format which lacks bond-order information (eg. prmtop). Loading from an sdf, for example, works fine.

**To Reproduce**
```
mol1 = sr.load('input.inpcrd','input.prmtop')[0]
rdmol1 = sr.convert.to_rdkit(mol1)
for atom in rdmol1.GetAtoms():
    print(atom.GetFormalCharge())

mol2 = sr.load('input.sdf')
rdmol1 = sr.convert.to_rdkit(mol2)
for atom in rdmol2.GetAtoms():
    print(atom.GetFormalCharge())
```

I'm not sure what the ideal solution would be, as the information about the correct structure isn't available. Possibly there could be some way to flag unlikely groups like carbanions?

[inputs.zip](https://github.com/user-attachments/files/22098202/inputs.zip)

lohedges:
This PR closes #350 by preferentially using `RDKit::determineBondOrders()` to infer bond orders when missing, only failing back to the existing (MDAnalysis-baed) inference routine when this fails. A unit test confirms that the problem inputs from the issue thread convert correctly when using SDF or AMBER inputs, i.e. the SMILES strings of the convert RDKit molecules match. All existing RDKit conversion tests pass, as do extended tests in BioSimSpace.

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -54,6 +54,8 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 
 * Added internal implementation of virtual sites in the Sire-to-OpenMM conversion layer 
 
+* Use ``RDKit::determineBondOrders()`` in Sire-to-RDKit conversion to infer bond orders.
+
 `2025.4.0 <https://github.com/openbiosim/sire/compare/2025.3.0...2025.4.0>`__ - February 2026
 ---------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


tom-potter-cresset (approved):
Thanks, this fixes the problem case I reported

PR #432, opened 2026-04-17
Fix out-of-bounds molidx searches silently returning the last molecule

### LINKED ISSUE #286: [BUG] Sire selection allows out-of-bounds indexing of molecules
Not sure if this is the intended behaviour or not, but it appears that out-of-bounds indexing using `MolIdx` in a Sire selection always gives the last molecule, e.g.:

```python
In [1]: import sire as sr

In [2]: mols = sr.load_test_files("ala.crd", "ala.top")


In [3]: mols.num_molecules()
Out[3]: 631

In [4]: mols["molidx 10000"]
Out[4]: Molecule( WAT:621 num_atoms=3 num_residues=1 )

In [5]: mols[-1]
Out[5]: Molecule( WAT:621 num_atoms=3 num_residues=1 )
```

This _isn't_ true for other indexing, e.g. atoms and residues:

```python
In [6]: mols.num_residues()
Out[6]: 633

In [7]: mols["residx 10000"]
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ in <module>:1                                                                                    │
│                                                                                                  │
│ /home/lester/.conda/envs/openbiosim/lib/python3.12/site-packages/sire/system/_system.py:75 in    │
│ __getitem__                                                                                      │
│                                                                                                  │
│     72 │   │   return self.__str__()                                                             │
│     73 │                                                                                         │
│     74 │   def __getitem__(self, key):                                                           │
│ ❱   75 │   │   return self.molecules()[key]                                                      │
│     76 │                                                                                         │
│     77 │   def __iadd__(self, molecules):                                                        │
│     78 │   │   self.add(molecules)                                                               │
│                                                                                                  │
│ /home/lester/.conda/envs/openbiosim/lib/python3.12/site-packages/sire/mol/__init__.py:505 in     │
│ __fixed__getitem__                                                                               │
│                                                                                                  │
│    502 │   │   │   # try to search for the object - this will raise                              │
│    503 │   │   │   # a SyntaxError if this is not a search term                                  │
│    504 │   │   │   # (and is instead a name)                                                     │
│ ❱  505 │   │   │   return __from_select_result(obj.search(key, map=map))                         │
│    506 │   │   except SyntaxError:                                                               │
│    507 │   │   │   # ignore SyntaxErrors as this is a name                                       │
│    508 │   │   │   pass                                                                          │
│                                                                                                  │
│ /home/lester/.conda/envs/openbiosim/lib/python3.12/site-packages/sire/mol/__init__.py:364 in     │
│ __from_select_result                                                                             │
│                                                                                                  │
│    361 │   │   )                                                                                 │
│    362 │                                                                                         │
│    363 │   if obj.list_count() == 0:                                                             │
│ ❱  364 │   │   raise KeyError("Nothing matched the search.")                                     │
│    365 │                                                                                         │
│    366 │   typ = obj.get_common_type()                                                           │
│    367                                                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
KeyError: 'Nothing matched the search.'
```

lohedges:
This PR closes #286, fixing a bug where out-of-bounds `molidx` searches silently returned the last molecule instead of raising a `KeyError`. The root cause was in `IDIndexEngine::searchMolIdx()`: unlike `searchIdx<T>` (used for atoms, residues, etc.), it had no single-value fast path and instead fell through to a `match()` call that used `slice.begin(count, true)` with `auto_fix=true`, silently clamping any out-of-bounds index to the last valid position. The fix adds a `_is_single_value` fast path to `searchMolIdx` that reads the raw index directly from `RangeValue::start` (bypassing the `_to_single_value` helper, which corrupts negative indices by mapping them against `INT_MAX`) and applies strict Python-style bounds checking. Negative indices within range (e.g. -1) continue to work correctly, while out-of-bounds values in either direction now return no match, consistent with residx and atomidx behaviour. A secondary bug in `MolIdx::map(const Molecules &)` where the loop counter i was never decremented is also fixed. A regression test is included.

Debugged with assistance from Claude Code Opus 4.6.

@chryswoods: Just want to address you concerns about different container types and use case. In the C++ layer we have `MolIdx::map(const Molecules&)`, where `Molecules` is a `QHash<MolNum, ViewsOfMol>`, so has no stable iteration order. I think this isn't an issue for user-facing code, since it goes through another pathway, but good to check. The comment referencing `static map()` in the issue thread, this pathway isn't touched by this fix, so was a red-herring, i.e. that's used for ranges and comparisons only. For the additional bug fix: I don't think this was ever an issue since this low-level overload probably wasn't used by any callers.

### DIFF
--- a/corelib/src/libs/SireMol/molid.cpp
+++ b/corelib/src/libs/SireMol/molid.cpp
@@ -266,6 +266,8 @@ QList<MolNum> MolIdx::map(const Molecules &molecules) const
             molnums.append(it.key());
             break;
         }
+
+        --i;
     }
 
     BOOST_ASSERT(not molnums.isEmpty());
--- a/corelib/src/libs/SireSearch/idengine.cpp
+++ b/corelib/src/libs/SireSearch/idengine.cpp
@@ -31,8 +31,8 @@
 #include "SireBase/booleanproperty.h"
 #include "SireBase/parallel.h"
 
-#include "SireMol/atomelements.h"
 #include "SireMol/atomcoords.h"
+#include "SireMol/atomelements.h"
 #include "SireMol/core.h"
 #include "SireMol/iswater.h"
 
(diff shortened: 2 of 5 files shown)


chryswoods (approved):
Thanks - painful, and a good fix. Thanks for finding this, especially as it fell through an assumption being incorrect when the match code is called from a container that expects (and has) bounds.

PR #442, opened 2026-05-12
Add softcore ring-breaking nonbonded force

lohedges:
This PR adds support for softcore ring-breaking and ring-making forces to smooth the onset of nonbonded forces when the ring topology changes, i.e. we don't immediately go from zero to full nonbonded, which can introduce a large repulsion energy. The code follows the same approach as the `ghost14` force field, only for the atoms involved in the ring-breaking bond. The required atoms are flagged during the BioSimSpace merge code and a unit test has been added to `somd2` to ensure that the new forces are present when they should be and are active/inactive at the correct points in the schedule. Putting these tests in `somd2` makes sure that we are testing the producing ring-breaking schedules, e.g. we don't have to duplicate and synchronise them here.

Testing on several complex perturbations has indicated that the new forces improve stability and overlap in the vicinity of the nonbonded onset. The major contributor to poor overlap is still the large conformational change (and resulting nonbonded energy change) as the ring topology adjusts, but these additions help as a stability safety net when the bond atoms get too close.

### DIFF
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -86,6 +86,8 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 
 * Add support for using a switching function for QM/MM simulations.
 
+* Add softcore ``CustomBondForce`` for ring-breaking and ring-making pairs.
+
 `2025.4.0 <https://github.com/openbiosim/sire/compare/2025.3.0...2025.4.0>`__ - February 2026
 ---------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


lohedges:
I've realised that I will also need to exclude one of the atoms in the bond pair from the `GhostNonGhostForce` if it is a ghost in one end state. This could happen in a perturbation where you add a dummy chain at one state, then bond it to an existing ring. As such, the atom at the end of the chain would also a ghost.

lohedges:
Actually, easier the other way round, i.e. don't create this force if a bond atom is a ghost.

PR #448, opened 2026-06-16
Fix issue #447

### LINKED ISSUE #447: [BUG] Hydrogens might not be detected depending on the HMR factor
Currently hydrogens might be detected by using a maximum mass of 2.5, which would fail for HMR factors above (roughly) 2.5. Currently the default in SOMD2 is 1.5, so this isn't an issue. However, for certain simulations, e.g. ring-breaks, it has been increased to 3, meaning that constraints aren't applied. We can either increase the default to something like 4, or add a map option, e.g. `max_h_mass`, that could be passed via SOMD2 depending on the HMR factor that is used.

### LINKED ISSUE #449: [BUG] ghost-14 force only created on the basis of the lambda=0 state
The condition for adding an entry to the `ghost-14` force is based on the lambda=0 states exception scale. This means that the force doesn't capture changes in connectivity, where additional/different handling is needed in the lambda=1 state.

### LINKED ISSUE #450: [BUG] Incorrect porting of SOMD1 hbonds-notperturbed
Currently, `h_bonds_not_heavy_perturbed` does _not_ reproduce SOMD1's `hbonds-notperturbed` in all cases. Specifically, the SOMD1 check for constraining atoms whose bond parameters don't change is _unconditional_, whereas in the logic used for SOMD2 it is performed only _after_ the light atom check based on the maximum end-state mass. This leads to invalid (or different) constraints for certain systems, which covers the problem cyclopropyl cases from Cresset.

lohedges:
This PR closes #447 by increasing the maximum mass used to detect hydrogen atoms along with exposing a map option to tune the value.

@chryswoods: In case I'm missing something: Is there a reason why 2.5 was chosen? This does work with what would be used for the default SOMD1/SOMD2 HMR factor, but not in general.

Also closes #449 by considering both end states when adding initial entries to the `ghost-14` force.

Also closes #450 by constraining unperturbed heavy atom bonds, regardless of mass, matching SOMD1.

### DIFF
--- a/corelib/src/libs/SireIO/biosimspace.cpp
+++ b/corelib/src/libs/SireIO/biosimspace.cpp
@@ -1227,6 +1227,16 @@ namespace SireIO
                 // Reduce the mass.
                 mass -= delta_mass;
 
+                if (mass.value() < 0)
+                {
+                    qWarning().noquote() << QObject::tr(
+                                                "Hydrogen mass repartitioning with factor %1 has resulted "
+                                                "in a negative mass for atom index %2. Consider using a "
+                                                "smaller repartitioning factor.")
+                                                .arg(factor)
+                                                .arg(idx.value());
+                }
+
                 // Set the new mass.
                 edit_mol = edit_mol.atom(idx).setProperty(mass_prop, mass).molecule();
 
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -88,6 +88,11 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 
 * Add softcore ``CustomBondForce`` for ring-breaking and ring-making pairs.
 
+* Add ``max_h_mass`` map option (default ``4.0`` g/mol) to control the mass threshold
+  used when identifying hydrogen atoms in the OpenMM conversion layer. Previously this
+  was hardcoded to ``2.5`` g/mol, which caused hydrogen constraints to be skipped when
+  an HMR factor greater than ~2.5 was applied.
+
 `2025.4.0 <https://github.com/openbiosim/sire/compare/2025.3.0...2025.4.0>`__ - February 2026
 ---------------------------------------------------------------------------------------------
 
--- a/doc/source/changelog.rst
+++ b/doc/source/changelog.rst
@@ -95,6 +95,8 @@ organisation on `GitHub <https://github.com/openbiosim/sire>`__.
 
 * Warn if hydrogen mass repartiting produces a negative heavy atom mass.
 
+* Update merge code to handle ``kartograf`` API changes in version 2.0.
+
 `2025.4.0 <https://github.com/openbiosim/sire/compare/2025.3.0...2025.4.0>`__ - February 2026
 ---------------------------------------------------------------------------------------------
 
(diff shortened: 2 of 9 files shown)


lohedges:
I bumped the threshold to 4.5 and set a max `h_mass_factor` in `SOMD2` of 4.0. This would also catch He, but that shouldn't be an issue for biomolecular simulations.

lohedges:
Actually, having thought more about this, a value of around 3.5 is safest. I've also updated the BioSimSpace and SOMD2 code to place an upper bound so that we don't end up with negative masses for heavy atoms.

lohedges:
I've also added a warning for situations where the HMR factor produces a negative heavy atom mass.

lohedges:
The tests appear to be failing due to a kartograph API change. I'll figure out what's changed and update. It might be that we need to add a pin.

lohedges:
This is actually a problem, since `SmallMolecueComponent` is now provided by `gufe`, which can cause lots of environment resolution issues, specifically the one used by Recursion. OpenFE have finally made `gufe` _optional_ for `lomap`, which was the problem with BioSimSpace, but have now made it required for `kartograf`, which would be pulled in by Sire. Bah.

Not sure what to do here. For now we could just build against the older version, or build against it but include as an optional dependency? I think the issue with the latter would be that it still affects the environment resolution in an undesirable way. However, by pinning to the old version I imagine that we would experience similar issues going forward, i.e. due to dependencies becoming outdated.

lohedges:
The merge tests are now failing due to energy differences, presumably caused by changes in the mapping:

```python
[38](https://github.com/OpenBioSim/sire/actions/runs/27674786376/job/81847337420#step:7:27739)
 │ >       assert nrg_merged_0.value() == pytest.approx(nrg_orig_merged_0.value(), abs=1e-3)
 │ E       assert -3.355915669284208 == -3.6911794314851085 ± 0.001
 │ E         
 │ E         comparison failed
 │ E         Obtained: -3.355915669284208
 │ E         Expected: -3.6911794314851085 ± 0.001
```

I'm not sure how much to investigate this, since other tests in the same function are commented out due to previous mismatches. (I guess that `orig_merged` no longer matches the current `merged`.)

lohedges:
Diffing atom_mapper.py between kartograf 1.2.0 and 2.0.0 there's one behavior-relevant change in `KartografAtomMapper.__init__`:

```python
  # kartograf 1.2.0
  map_hydrogens_on_hydrogens_only: bool = False

  # kartograf 2.0.0
  map_hydrogens_on_hydrogens_only: bool = True
```

This explains the error, which is due to a difference in the mappings between versions:

  - kartograf 1.2: {1: 0, 3: 1, 11: 2, 12: 3, 13: 4} — 5 pairs, including carbon idx 1 mapped onto methane's hydrogen
  idx 0.
  - kartograf 2.0: {3: 1, 11: 2, 12: 3, 13: 4} — 4 pairs, the carbon→hydrogen pair is gone.

In 1.2, with `map_hydrogens_on_hydrogens_only=False`, the hydrogen-mapping pass was allowed to match a heavy atom (a shrinking methyl carbon) onto a real hydrogen position. In 2.0 this default flipped to True, so heavy atoms can no longer be matched to hydrogens during that pass, so one fewer atom is mapped, which changes the merged topology and therefore the potential energy at λ=0/1, causing the test assertion to fail.

We can fix this be explicityly setting `map_hydrogens_on_hydgens_only=False` when creating the `KartografAtomMappper`, which avoids the change in default between versions.

lohedges:
Despite the CI passing yesterday we are now seeing a repeated test failure for Linux Python 3.11 for a completely unrelated test (AMBER CMAP parsing). This is reproducible across CI runs. Will investigate on Monday.

PR #459, opened 2026-08-04
Reimplement lazy importing using importlib

lohedges:
This PR switches to using our own internal, native Python lazy import system. This fixes several issues with the third party package that we were using:

- Fixed spurious isinstance()/pickling failures across process boundaries. This was the original production bug that started this work: lazily-loaded modules could end up as two distinct class objects for the same module path when independently loaded in separate worker processes (joblib/multiprocessing), breaking type checks on objects passed between them.
- Fixed the same duplicate-object bug for direct submodule imports reached before the parent package had been touched (e.g. via import pkg.sub.leaf, or pickle resolving a class's __module__).
- Silent corruption when third-party code introspects sys.modules. Encountered via openff/nagl: PyTorch's custom_op registration calls inspect.getmodule(), which checks hasattr(module, '__file__') on every entry in sys.modules — under lazy_import, that forced every lazily-loaded module in the process to load all at once, mid-import of unrelated code, with no error raised. Not a crash, but a silently wrong result (an empty/corrupted return value, NAGL charge assignment silently disabled) that's much harder to catch than a clean failure.

This also removes a GPLv3 dependency from our stack.

Written and debugged with help from Claude.


lohedges:
CI failures are caused by actions flakiness and intermittent I/O and connection glitches.

PR #484, opened 2026-09-21
Fix issue #483

### LINKED ISSUE #483: [BUG] Dynamics.run() steps in 50 step blocks and evaluates the energy after each one
`Dynamics.run()` drives the integrator in fixed blocks of 50 steps, and after every block `_exit_dynamics_block()` calls `getState(getEnergy=True)`, which forces a full force evaluation and a GPU sync even when nothing is being saved; the state is used only for the simulation time and the kinetic energy blow-up check. On a 46646 atom protein-ligand system at 4 fs on an RTX 5070 Ti this costs about 0.85 ms per block against 17.6 ms of stepping: the raw integrator runs at 0.363 ms/step and `Dynamics.run()` with no energies or frames at 0.380 ms/step, a 4.7% loss, with the energy scan and frame saving adding nothing further at their usual frequencies. The block size does not contribute to crash recovery, which restores the `_pre_run_state` snapshot taken before `run()` was called and re-minimises, so it only serves the progress bar and the timeout check. The block size should instead be set by what is being recorded, i.e. the number of steps to the next energy or frame save (or the end of the run), falling back to the current 50 steps only when a progress bar is shown. The per-block `getState(getEnergy=True)` can stay, since the blow-up check needs the kinetic energy and OpenMM has no way to request that without the potential; at 250 step blocks it costs about 0.3%.

lohedges:
This PR closes #483. `Dynamics.run()` stepped the integrator in fixed blocks of 50 steps, and after each block `_exit_dynamics_block()` calls `getState(getEnergy=True)`, which is a full force evaluation and a GPU sync. The blocks only exist to update the progress bar; crash recovery is unaffected by them, since it restores the snapshot taken before `run()` was called. Each block now runs to the next frame or energy save, or the end of the run, and the 50 step cap is applied only when the progress bar is visible. `ProgressBar` gains an `is_silent()` query for this, tracked in the Python wrapper alongside `set_silent()`. Because a block can now be long, its length is fixed from both the next frame and the next energy save before either flag is set, so that one save shortening the block cannot move the other; a side effect is that saves now land on the requested step rather than at the end of a 50 step block. The existing dynamics tests cover the frame and energy scheduling and pass with the progress bar both shown and silenced.

On a 46646 atom protein-ligand system at 4 fs on an RTX 5070 Ti, with the progress bar silenced as somd2 does, `Dynamics.run()` with nothing saved went from 0.380 to 0.360 ms/step against 0.356 ms/step for the raw integrator, and with an 11 window energy scan every 2 ps from 0.381 to 0.367 ms/step.


chryswoods (approved):
👍

lohedges:
Failures are GitHub actions network issues causing IO errors.

PR #488, opened 2026-09-23
Fix issue #487

### LINKED ISSUE #487: [BUG] SireOpenMM wrapper generation drops SireBase::Property base class
When the SireOpenMM Python wrappers are regenerated, the scripts leave SireBase::Property out of the base classes of the QM engine classes (QMEngine, NullQMEngine, PyQMEngine and TorchQMEngine). SireOpenMM has its own header-scanning script, and it only records the module's own classes. The wrapper generator only adds base classes it knows about, so it skips Property. Up to now PyQMEngine and TorchQMEngine have been fixed by hand after each regeneration, but a full rebuild would silently undo that. QMEngine and NullQMEngine are already missing the base class. The SireOpenMM scanning script should be updated so it knows about the core Sire classes, and the affected wrappers should be corrected.

lohedges:
Regenerating the SireOpenMM Python wrappers left out the link between the QM engines and Sire's core Property class. Up to now we've patched some of the wrappers by hand after each regeneration, and a full rebuild could quietly undo that. This PR fixes the regeneration scripts so the link is always kept, and it corrects the wrappers that were already missing it, including the lambda lever and perturbable molecule classes. A new test checks that these objects can be stored and retrieved as properties, which is how Sire passes the QM engine around.

Closes #487.


lohedges:
The PR also resolves an issue caused by https://github.com/OpenBioSim/biosimspace/issues/556. This change pulled in PyTorch, which triggered the `TorchQMForce` build on Windows, which failed due to C++20 headers. I also took the time to tidy up some issues related to compiler warnings, which exposed a few minor bugs along the way.

